Issue #16, opened 2018-02-23
Gotham frontend


ParisLiakos:
Feature
=========

Gotham is probably the best solution for web development since iron, nickel are no longer actively maintained and rocket requires nightly.
We should build against 0.2 version, so postponing until its released.

--------

 - [x] A pull request (#18)

HeroicKatora:
Another future based frontend, that sounds very nice. As a basic orientation, looking at the implementation of [`actix`](https://github.com/HeroicKatora/oxide-auth/blob/actix/src/frontends/actix.rs) may be of interest.

ParisLiakos:
Thank you! The actix implementation helped a lot :+1: 
I will open a pull request soon for this, need to do some cleanup first.

PR #18, opened 2018-03-25
Add a gotham frontend

ParisLiakos:
This adds a gotham frontend and an example for using it.

It is not added in the default features because the support module/dummy client depends on a struct which in turn depends on macros from other crates :) 
If you have any idea how to accomplish this without littering the other examples with unneeded extern crate calls, let me know.

I am not sure whether everything makes sense or not..Any feedback/suggestion is appreciated!

 - [x] This change has minimal documentation
 - [x] Corresponds to issue #16

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,7 +34,7 @@ urlencoded = { version = "^0.6", optional = true }
 lazy_static = { version = "1.0", optional = true }
 
 [features]
-default = ["gotham-frontend", "iron-frontend", "rouille-frontend"]
+default = ["iron-frontend", "rouille-frontend"]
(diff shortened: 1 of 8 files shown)


HeroicKatora:
Hm, maybe `support` should be split up anyways. The problem is only going to get worse, so if you could move your support mod into `support-gotham` instead, that would allow it to be enabled by default. Its purpose is mainly to cleanly separate the necessary server code from the dummy client module, so there is no need for everything to be clumped together.

ParisLiakos:
Thanks for the quick review! 
Ok, i did that, no problems now :+1:
Gotham feature is in the defaults now

Issue #15, opened 2018-02-22
Actix frontend


HeroicKatora:
Feature
=========

Actix is arguably a quickly evolving library and its web framework has been proven to be very performant.

The greatest challenge is translation between `Future` structures and the sequential handling of requests here.

Tracking pull request
--------

 - [x] The development branch has been merged into dev-v0.4.0

HeroicKatora:
This will not land in master for now because of the multiple native library dependencies. Rouille and Actix collide with different versions of [`brotli2`] (actix using `v0.3.2` while rouille uses `v0.2.2`). This would best be solved by refactoring this into a workspace instead. 

The other problem is ring, used by `cookie` as well as this library itself. For more informationg and tracking on updates, see the proposed [resolution to ring linking issue](https://github.com/briansmith/ring/pull/619).

[`brotli2`]: https://crates.io/crates/brotli2

HeroicKatora:
It sometimes seems as if a request was blocking and leading to a timeout on all future requests. I have no idea why that is?

trsh:
Want this badly. Need help?

HeroicKatora:
After reworking the frontend to provide an actor based endpoint, everything now works as expected. This means that the basic architecture for this frontend is likely staying as is although the interfaces are not stable yet.

joshsharp:
Hi! I'm building a backend in Actix that'll use OAuth2 and Oxide-auth looks like it'd fit the bill perfectly. I wondered if I could ask, if it's not too pushy to do so, what the status of this frontend is. Is there more left to do, and/or can I get Cargo to use the dev branch so I can take it for a spin? Thanks!

HeroicKatora:
You have two choices:

* Use the `0.4.0-preview.0` version, which includes actix enabled by default (you may want to disable all other frontends).
* Override the dependency in Cargo.toml and point it to the `dev-v0.4.0` branch:

```
[dependencies]
oxide-auth = { git = 'https://github.com/HeroicKatora/oxide-auth', branch = 'dev-v0.4.0' }
```

The first option will give you a  (hopefully) stable experience while the latter will give you new interfaces sooner. Stable version v0.4.0 should be selected, once published, with both version by using `cargo update`. 

Status on the development is that I'm busy with University end terms until the end of the month and I will continue to work towards the release of 0.4 in August.

joshsharp:
Excellent, thank you! I'll go with the former I think.

zoidbergwill:
[...]

PR #21, opened 2018-11-13
Fix compilation error

subhojit777:
This changes/fixes/improves #15;

 - [x] Corresponds to issue (#15 )

### DIFF
--- a/examples/authorization_actix.rs
+++ b/examples/authorization_actix.rs
@@ -43,15 +43,15 @@ pub fn main() {
     // Emulate static initialization for complex type
     let scopes: &'static _ = Box::leak(scopes);
 
-    let endpoint: Addr<_> = CodeGrantEndpoint::new((clients, authorizer, issuer))
-        .with_authorization(|&mut (ref client, ref mut authorizer, _)| {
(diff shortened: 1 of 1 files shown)


HeroicKatora:
This seems to be fine for me, does the compilation error happen on nightly?

subhojit777:
`rustc 1.30.1 (1433507eb 2018-11-07)`

HeroicKatora:
Weird, I can't replicate it with exactly the same compiler version. 

Maybe your work is based on some intermediate commit in a longer commit series. Can you try to base your work on the tag `v0.4.0-preview.0`? Note that the development branch is expected to change a lot, including breaking changes and non-compiling versions but only all tagged versions (aka. those published to <crates.io>) should work correctly. Still, your patch seems to be based on commit e6211cb855395f4d52c7c4daed451fab06e68398 which I can compile fine.

subhojit777:
I was not able to compile with the `0.4.0-preview.0` version. There is a conflict with `ring` crate - https://github.com/HeroicKatora/oxide-auth/issues/15#issuecomment-435942381

HeroicKatora:
Sorry that I never got around to this as it was a preview version and I could reproduce it. This should not happen in `v0.4` release.

Issue #29, opened 2019-05-28
Partial-normalization of redirect_uri


Nemo157:
# Bug report

Causing unexpected behaviour

The `Client` constructors require submitting a `Url` object for the `redirect_uri`, while the value submitted as part of an `access_token` request is kept as just a string. This can result in the URI being normalized when registering a client, but not when attempting to validate an `access_token` request against the registered client.

## Reproduction

 * Register a `Client` in a `ClientMap` using a `redirect_uri` generated from `"foobar://oauth2redirect".parse().unwrap()`.
 * Attempt to generate an access token from a request containing valid properties including `"redirect_uri": "foobar://oauth2redirect"`
 * An `invalid_grant` error is returned

Adding in some debugging shows that this is failing because the recorded `redirect_uri` is having an additional `/` appended to it when compared at https://github.com/HeroicKatora/oxide-auth/blob/d625aea571efe7d21a2a9f6644feb41643158f6c/src/code_grant/accesstoken.rs#L128

```rust
[/Users/nemo157/.cargo/registry/src/github.com-1ecc6299db9ec823/oxide-auth-0.4.2/src/code_grant/accesstoken.rs:128] (saved_params.client_id.as_ref(), saved_params.redirect_uri.as_str()) = (
    "foobar",
    "foobar://oauth2redirect/",
)
[/Users/nemo157/.cargo/registry/src/github.com-1ecc6299db9ec823/oxide-auth-0.4.2/src/code_grant/accesstoken.rs:128] (client_id, redirect_uri) = (
    "foobar",
    "foobar://oauth2redirect",
)
```

## Expected behaviour

The access token request to succeed

Tracking pull request
--------

 - [ ] A pull request (does not yet exist/exists at #&#8230;)

Nemo157:
Just to make sure I checked RFC 3986 to ensure path-less URIs are valid, [section 3.3](https://tools.ietf.org/html/rfc3986#section-3.3):

> If a URI contains an authority component, then the path component must either be empty or begin with a slash ("/") character.

HeroicKatora:
The tight check is motivated by [section 3.1.2.3](https://tools.ietf.org/html/rfc6749#section-3.1.2.3) which states:

> If the client registration included the full
   redirection URI, the authorization server MUST compare the two URIs
   using simple string comparison as defined in [RFC3986] Section 6.2.1.

[...]

PR #30, opened 2019-06-02
Check redirect equivalence as url instead of str

HeroicKatora:
This fixes an issue where the information lost from parsing the client
provided redirect uri string as an url would lead to unexpected
behaviour when retrieving the access token for the client. Fixed by
performing a full url based comparison.

 - [x] Corresponds to issue #29


HeroicKatora:
Superseeded by #32 because I was unsure how to get CI to pick this up. Turns out you just need a configuration but it felt stupid to reopen this one instead.

PR #32, opened 2019-06-02
Check redirect equivalence as url instead of str

HeroicKatora:
This fixes an issue where the information lost from parsing the client
provided redirect uri string as an url would lead to unexpected
behaviour when retrieving the access token for the client. Fixed by
performing a full url based comparison.

 - [x] Closes: #29


HeroicKatora:
Hm weird, can't get cirrus ci to pick it up. Probably because there is no configuration for it? I'll try rebasing it.

Issue #39, opened 2019-08-08
Extensions::iter_private returns PublicExtensions?


hrbigelow:
# Bug report

In [this](https://github.com/HeroicKatora/oxide-auth/blob/9e546ce230fb7c207835dd7415a792ca358e4cd2/src/primitives/grant.rs#L139) code (below):
```
    /// Iterate of the public extensions whose presence and content is not secret.
    pub fn iter_public(&self) -> PublicExtensions {
        PublicExtensions(self.extensions.iter())
    }

    /// Iterate of the private extensions whose presence and content must not be revealed.
    pub fn iter_private(&self) -> PublicExtensions {
        PublicExtensions(self.extensions.iter())
    }
```
shouldn't it be `PrivateExtensions`?

HeroicKatora:
Oh my. Yes. How did it take so long to actually get noticed? Well, at least the fix is straightfoward.

hrbigelow:
No worries - I think writing a test suite for auth is really hard, given the multi-server nature of it.  The code is super clean in any case!

hrbigelow:
I'm studying your code so I can better understand how OAuth works, (and also use it correctly to implement an email-based signup flow)

One other bug FYI:  "ephermal" should be "ephemeral" at src/primitives/generator.rs:129

HeroicKatora:
@hrbigelow I even have the correct spelling [in issuer](https://github.com/HeroicKatora/oxide-auth/blob/f91df51aa4e18e94d262046051ea17367366ed29/src/primitives/issuer.rs#L181) :smile:

HeroicKatora:
Appreciate all of the reviews and nits by the way.

PR #40, opened 2019-08-08
Soft fixes for broken interfaces

HeroicKatora:
Fixes the interface issues discovered in #39.

Deprecates the broken interface, they will be removed in the next major version. We avoid a breaking change in the case of private extension iteration by introducing a boolean flag to turn the public extension iterator into the private one. Renames the constructors for both iterators to a more 'rusty' version which is more similar to `HashMap::keys` and other specialized container iterators.

This fixes a bug where the `Assertion` grant type could silently discard private extensions instead of returning with an error.

 - [x] This change has tests
 - [x] Corresponds to issue #39

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -22,6 +22,7 @@ nightly_task:
   cargo_cache:
     folder: $CARGO_HOME/registry
     fingerprint_script: cat Cargo.lock
+  pre_script: cargo --version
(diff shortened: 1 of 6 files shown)


HeroicKatora:
This fails due to nightly breakage, that would only affect testability of a dependency of `iron`, not testability of `oxide-auth`. See [this PR](https://github.com/conduit-rust/route-recognizer.rs/pull/25) that is available to fix it.

Issue #47, opened 2019-08-18
Split crate per web framework


HeroicKatora:
# Project Improvement

The crate should be split into a core crate and one binding crate per web framework.

Several times, I have seen a question to the effect of: 'does `oxide-auth` already support *X*`? And it was not seemingly not conveyed in the crate structure that the bindings could be added in an external crate. That would be clearer if it were already the case for 'officially' support web servers.

Contributions can focus on one specific web server without worrying about unrelated code.

Improvements or restructings in one webserver don't require major releases for all others.

No more features to activate. This reduces compile time (and thus testing) and friction due to less dependencies by default. There would no longer a problem of a default endorsed crate, which may change rather quickly in the Rust ecosystem and should not require releases at all.

## New structure

The structure, as reverse dependencies, would look like this:

```
oxide-auth
|- oxide-auth-actix
|- oxide-auth-rouille
|- oxide-auth-iron
|- oxide-auth-rocket
\- oxide-auth-tide (???)
```

## Other context

Upgrading `actix` will be a potentially large undertaking involving reworking actors to futures. This should not slow the progress of other web server crates, including new ones such as tide. 

# Tracking pull request

 - [x] A pull request exists at #50 (merged)

PR #49, opened 2019-08-23
Split oxide-auth into crates for each integration

asonix:
Start splitting each webframework into a subcrate, exposing each integration through feature flags on a facade crate.

Currently features
- core crate containing the basic types and traits
- actix 0.7 integration crate
- iron 0.6 integration crate
- rocket 0.4 integration crate
- rouille 3.0 integration crate

notes
 - [x] Corresponds to issue #47 


asonix:
there's probably a lot more changes in this PR than there needed to be since my editor automatically cargo-formats each file I edit on save...

asonix:
there's probably a lot more changes in this PR than there needed to be since my editor automatically cargo-formats each file I edit on save...

asonix:
I think this is ready to be looked at, now

HeroicKatora:
Just wow, thank you so much. There is already such a split in progress, see [the `next` branch](https://github.com/HeroicKatora/oxide-auth/tree/next). Can you rebase your frontend crates/implementions onto that? Examples should be placed next to their integration crates, e.g. in `oxide-auth-rouille/examples` for rouillle examples.

You seem to have put some extra work into keeping the central interface of the main crate in-tact with features, I've entertained directing users into choosing the correct frontend crate instead. The features have been a pain to work with in the past, especially since choosing two different frontends is seldomly useful but extremely hard to maintain. And the major version number would not be anywhere near as accurate as useful. So maybe `oxide-auth 0.5` should just be `oxide-auth-core` with `oxide-auth-actix` being the dependency to choose when coding for `actix`.

> there's probably a lot more changes in this PR than there needed to be since my editor automatically cargo-formats each file I edit on save...

Highly preferrably without editor autoformatting, it messes with git a log. For example `examples/iron-example/src/main.rs` is not even recognized as a moved file anymore... If it has to be, then first do all the autoformatting in one streak of commits and then the changes themselves. I've already had a hard time picking out the places where `feature` was still in use for example.

asonix:
Changes ported over to #50

PR #50, opened 2019-08-28
Port in subcrate updates from asonix/subcrates

asonix:
Currently ported:
- [x] oxide-auth-actix
- [x] oxide-auth-iron
- [x] oxide-auth-rocket
- [x] oxide-auth-actix/examples
- [x] oxide-auth-iron/examples
- [x] oxide-auth-rocket/examples

PR Status:
 - [x] Corresponds to issue #47 


asonix:
@HeroicKatora Here's a PR based of of the `next` branch

oxide-auth-rocket/src/lib.rs
```diff
@@ -131,49 +132,85 @@ impl<'r> WebRequest for OAuthRequest<'r> {
...
+    fn from_request(request: &'a Request<'r>) -> Outcome<Self, (Status, Self::Error), ()> {
+        Outcome::Success(Self::new(request))
+    }
+}
+
+/// Response type for Rocket OAuth requests
+///
+/// A simple wrapper type around a simple `rocket::Response<'r>` that implements `WebResponse`.
+#[derive(Debug)]
+pub struct OAuthResponse<'r>(pub Response<'r>);
```
HeroicKatora: Preferred ordering for closely related components is

1. All struct definitions
2. Inherent methods and core trait impls
3. 'Helper' traits such as `Default`, `From`, ...

Please move the struct definition to below `OAuthRequest` and the `Default` and `From` impl under `WebResponse`.

oxide-auth-iron/src/lib.rs
```diff
@@ -0,0 +1,173 @@
...
+            EndpointError::DenySilently => Status::BadRequest,
+            EndpointError::PrimitiveError => Status::InternalServerError,
+        };
+
+        OAuthError(IronError::new(as_oauth, status))
+    }
+}
+
+#[derive(Debug)]
+pub struct OAuthResponse(pub Response);
```
HeroicKatora: Please reorder like the other. Also, this is missing a doc-comment.

 It would be great if there was `#![warn(missing_docs)]` in all individual libraries.

oxide-auth-iron/src/lib.rs
```diff
@@ -0,0 +1,173 @@
...
+}
+
+impl Into<Response> for OAuthResponse {
+    fn into(self) -> Response {
+        self.0
+    }
+}
+
+#[derive(Debug)]
+pub struct OAuthError(pub IronError);
```
HeroicKatora: Same here.

oxide-auth-iron/src/lib.rs
```diff
@@ -0,0 +1,173 @@
...
+use url::Url;
+
+/// Errors while decoding requests.
+pub enum Error { 
+    /// Generally describes a malformed request.
+    BadRequest,
+}
+
+#[derive(Debug)]
+pub struct OAuthRequest<'a, 'b, 'c: 'b>(pub &'a mut Request<'b, 'c>);
```
HeroicKatora: Doc-comment. See `rocket` for a rough template for these all.

HeroicKatora:
Something that is slightly unwieldy to discuss in a review thread: The encapsulation of public member in a public tuple type as `pub struct OAuthRequest(pub server_lib::Request)` is a choice that may be worth additional consideration. *And I'm not personally sure what is the best choice*.

I would have opted for an opaque type, which is the route I went for `rouille`, so that the inner representation can be changed. This is coupled with a `Deref` implementation on the request type which makes it unecessary to explicitely spell out `request.0.` when inspecting the request in consent solicitor function for example. The Rust API guidelines would usually only advise `Deref` on pointer but this may be a comparable cause.

However, it means there should not be *any* inherent methods with `&self` parameter. This should mostly be fine when having enough `From` and `Into` implementations, as well as the non-inherent `::new(server_lib::Request)` but it may lead to more awkward code as a result.

asonix:
I'm hesitant to impl `Deref` on a wrapper type, since that seems basically the same as exposing `wrapper.0` to me. If the internals were ever to change to not wrap that type, the `Deref` impl would be invalid. Maintaining `From` and `Into` would be less difficult, and providing methods on the wrapper type to abstract over the internals makes more sense to me.

HeroicKatora:
> I'm hesitant to impl Deref on a wrapper type, since that seems basically the same as exposing wrapper.0 to me.

Not quite. `struct Wrapper(pub Member)` exposes that we *only* wrap `Member` and in which form (can't reformat it to a proper struct) but provides a magic constructor. On the other hand, `Deref<Target=Member>` exposes that we wrap *some* `Member` but it would allow adding extra members besides it or restructure the internals of where that `Member` is placed (could even be allocated for example to reduce struct size itself).

But as long as `From` and `Into` work, I agree it seems more sensible not to have these impls, definitely also from a maintenance perspective.

So go with the style you did in `iron`, it seems somewhat clean. For the case of whether to expose `get(&self) -> server_lib::Request`, I think that could also be decided on a per-library basis. It is unecessary for Rocket where the request is a simple reference and thus `Copy` anyways.

HeroicKatora:
Thank you so much, again! This closes the main issue, anything else that was discussed can and should be done as a separate issue when/if it comes up.

Issue #37, opened 2019-06-29
actix-web update to 1.x


s-panferov:
# Project Improvement

`actix-web` 1.x has some breaking changes. The library and examples need to be updated to remain useful.

# Tracking pull request

 - [ ] A pull request (does not yet exist)

asonix:
Here's something interesting I just realized: Actix Web 1.0 with the `secure-cookies` feature depends on ring 0.14, which is incompatible with ring 0.13... I'm not sure how to resolve this

HeroicKatora:
Potential ways:

* Have multiple features that allow selecting a compatible version of `ring` in `oxide-auth`. Finally somewhat sane since frontends can each select their exclusive features.
* Remove the dependency on `ring`. This would be a very slightly breaking change as the HMAC key can currently be converted from a `ring` struct. The potential of `ring` is barely used so this doesn't require looking for another general purpose crypto library. But needs some auditing of that library, I believe the others are not as popular.

asonix:
```toml
ring = { version = "0.14", optional = true }
ring13 = { version = "0.1.0", optional = true }
```

Using a ring shim crate doesn't work, even if both `ring` and `ring13` are optional dependencies.

HeroicKatora:
It seems that all crates in a workspace need to lead to a single selection of packages and version, and linking collisions from native libraries are also checked across all these packages. Otherwise it would work to specify manually a SemVer range like so:

> ring = ">=0.13,<0.17"

Maybe we can 'just' get rid of the workspace and build each separately.

HeroicKatora:
Hm, `0.14` has a breaking change to https://docs.rs/ring/0.14.0/ring/pbkdf2/fn.verify.html which is in use. This is not so simple, needs to be features.

HeroicKatora:
Cheeky idea: turn `iterations` into a `NonZeroU32` and use `self.iterations.into()`. This should have the correct behaviour in all versions. Once it is a no-op and for `0.13` it selects `From<NonZeroU32> for u32`.

asonix:
I've gotten rid of the workspace entirely on my box and i'm still not able to compile a package that optionally depends on ring13 and ring14

HeroicKatora:
Cargo doesn't seem to like it, can't have conflicting optional dependencies it seems. The whole workspace needs to resolve to the same ring version though, as well. Maybe a separate directory just for testing both/all possible ring versions should be added outside the workspace. Then at least the current setup can stay.

[...]

PR #51, opened 2019-08-29
Bump actix-web to 1.0

asonix:
 - [x] This change has tests (does the example count?)
 - [x] Corresponds to issue #37 


HeroicKatora:
This removes any possibility of `async`/message passing code but I guess `actix` 1.0 support is more important. I'll have to sleep about it but am inclined to accept that since actual `std::future::Future` compatibilty requires changes to the underlying library anyways and would get us a much *better* interface for new actix as well (right?).

HeroicKatora:
>  This change has tests (does the example count?)

Yes :)

asonix:
Actix 1.0 is still using futures 0.1, but it is fairly simple to `.compat()` between 0.1 and std futures.

asonix:
It might also be worth putting the state in an actor instead of behind mutexes in the example, since then that actor would have ownership and would be able to run the necessary operations without `web::block`.

asonix:
I went ahead and built an actor to serialize the oxide operations without using a mutex. This is also closer to the "right way" to do this in production systems.

HeroicKatora:
A note on style: The 'better way'™ for production would be to implement [`oxide_auth::endpoint::Endpoint`](https://docs.rs/oxide-auth/0.4.5/oxide_auth/endpoint/trait.Endpoint.html) for the Actor containing all the state, instead of doing this dance with `Generic` that is transformed multiple times.

But more importantly, I think the example *looks* bloated. If I were to see the library for the first time, ~350 lines across multiple files for a very simple oauth-flow that doesn't protect any useful resource is .. somewhat daunting.

Here is a thought to maybe simplify this, and integrate `actix` more closely:

If the `State` were already a full `Endpoint` then each implementation of `OxideOperation` would be straightforward:

1. Try to construct the Flow
2. Execute the flow

Note that this implementation could also be generic over `Endpoint`. This is very similar to the `message` module that was previously in place there, but only requires the user to add the *one* blanket impl for `actix::Handler<OxideMessage<T>>` that is in the example. This would be nicer than the previous version, which required the actor to implement all separately and would get rid of `AsActor` accordingly.

Thus, maybe we can move `OxideOperation` into the library in the following form:

```
pub trait OxideOperation: Sized + 'static {
    type Item;
    type Error: std::fmt::Debug;

    fn run(self, state: impl Endpoint<OAuthRequest>) -> Result<Self::Item, Self::Error>;

    fn wrap(self) -> OxideMessage<Self> {
        OxideMessage(self)
    }
}

pub struct OxideMessage<T>(T)
where
    T: OxideOperation;

impl<T> Message for OxideMessage<T>
where
    T: OxideOperation + 'static,
{
    type Result = Result<T::Item, T::Error>;
}
```

That would require the user only to implement the `Actor` and `Handler` impls as follows:

```
impl Actor for State {
    type Context = Context<Self>;
}

impl<T> Handler<OxideMessage<T>> for State
where
    T: OxideOperation + 'static,
{
    type Result = Result<T::Item, T::Error>;

    fn handle(&mut self, OxideMessage(op): OxideMessage<T>, _: &mut Self::Context)
        -> Self::Result
    {
        // Where `as_endpoint` gets a reference to the contained endpoint.
        op.run(self.as_endpoint())
    }
}
```

HeroicKatora:
Looks much better. I'm going to do a full review tomorrow (eta), in the meantime you can add yourself to the credited [Contributors.txt](https://github.com/HeroicKatora/oxide-auth/blob/master/Contributors.txt) with as much or few metadata as you feel comfortable with.

asonix:
Since `OAuthRequest` consumes the request body on extraction, I've added `OAuthResource` which has `.into_request()` for guarding resources without consuming the body.

oxide-auth-actix/src/lib.rs
```diff
@@ -4,69 +4,457 @@
...
-pub use oxide_auth::endpoint::{PreGrant, OAuthError, OwnerConsent, OwnerSolicitor};
-pub use oxide_auth::primitives::grant::Grant;
-pub use self::request::OAuthFuture;
-pub use self::request::OAuthRequest;
-pub use self::request::OAuthResponse;
+/// Describes an operation that can be performed in the presence of an `Endpoint`
+///
+/// This trait can be implemented by any type, but is very useful in Actor scenarios, where an
+/// Actor can provide an endpoint to an operation sent as a message.
+pub trait OxideOperation: Sized + 'static {
```
HeroicKatora: Should be `OAuthOperation` for consistency with the other names.

HeroicKatora: Or did you intend to have two different naming schemes?

oxide-auth-actix/src/lib.rs
```diff
@@ -4,69 +4,457 @@
...
-// pub use self::endpoint::CodeGrantEndpoint;
-pub use oxide_auth::endpoint::{PreGrant, OAuthError, OwnerConsent, OwnerSolicitor};
-pub use oxide_auth::primitives::grant::Grant;
-pub use self::request::OAuthFuture;
-pub use self::request::OAuthRequest;
-pub use self::request::OAuthResponse;
+/// Describes an operation that can be performed in the presence of an `Endpoint`
+///
+/// This trait can be implemented by any type, but is very useful in Actor scenarios, where an
+/// Actor can provide an endpoint to an operation sent as a message.
```
HeroicKatora: The documentation here should include the usage from the example. It should be clear that this trait exists to easily turn a user-provided actor into a handler for OAuth messages by a simple 

> `impl<T: OxideOperation> Handler<OxideMessage<T>> for State` { ... }

oxide-auth-actix/examples/actix-example/src/main.rs
```diff
@@ -0,0 +1,241 @@
...
+    }
+}
+
+impl Actor for State {
+    type Context = Context<Self>;
+}
+
+impl<T> Handler<OxideMessage<T>> for State
+where
+    T: OxideOperation + 'static,
```
HeroicKatora: `OxideOperation` already asserts `'static`. Consider dropping this one to make the example terser.

oxide-auth-actix/examples/actix-example/src/main.rs
```diff
@@ -0,0 +1,241 @@
...
+    T::Item: 'static,
+    T::Error: 'static,
```
HeroicKatora: Same for `OxideOperation::{Item, Error}` that are already declared `'static`.

oxide-auth-actix/src/operations.rs
```diff
@@ -0,0 +1,78 @@
+use crate::{OAuthRequest, OAuthResponse, OxideOperation, WebError};
+use oxide_auth::{
+    endpoint::{AccessTokenFlow, AuthorizationFlow, Endpoint, RefreshFlow, ResourceFlow},
+    primitives::grant::Grant,
+};
+
+/// Authorization-related operations
+pub struct Authorize(pub OAuthRequest);
```
HeroicKatora: These could be polymorphic:

`struct Authorize<Req = OAuthRequest>(pub Req)` and then conditionally implement `OxideOperation` when the type, the error type and response type fulfill the required trait bounds from `OxideOperation`.

Note that, in production, it is useful to wrap a `OAuthRequest` into a custom `impl WebRequest` to enrich it with additional information from the original http request that is available in the solicitor for example.

oxide-auth-actix/src/lib.rs
```diff
@@ -4,69 +4,457 @@
...
+            .insert(CONTENT_TYPE, HttpTryFrom::try_from("application/json")?);
+        Ok(())
+    }
+}
+
+impl<T> Message for OxideMessage<T>
+where
+    T: OxideOperation + 'static,
+    T::Item: 'static,
+    T::Error: 'static,
```
HeroicKatora: These `'static` bounds also implied by `OxideOperation`.

oxide-auth-actix/src/operations.rs
HeroicKatora: Also, adding generic parameters need not be the correct solution. It's just something that isn't quite clear to me.

asonix: I've updated the Message type to carry a generic "extras" parameter which can be used in implementations to send data long with the OAuthRequest or ignored entirely for simple cases.

HeroicKatora: @asonix Resolving this since the `Extra` parameter on the message alone seems clearer!

oxide-auth-actix/src/lib.rs
asonix: That wasn't intentional. I've migrated everything to OAuth[Thing]

oxide-auth-actix/examples/actix-example/src/main.rs
```diff
@@ -0,0 +1,241 @@
...
+    // Create the main server instance
+    HttpServer::new(move || {
+        App::new()
+            .data(state.clone())
+            .wrap(Logger::default())
+            .service(
+                web::resource("/authorize")
+                    .route(web::get().to_async(
+                        |(mut req, state): (OAuthRequest, web::Data<Addr<State>>)| {
+                            auth_in_progress(&mut req);
```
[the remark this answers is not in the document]
asonix: This has been removed now that the example sends FnSolicitors for evaluation

oxide-auth-actix/examples/actix-example/src/main.rs
```diff
@@ -0,0 +1,236 @@
...
+                                            OAuthResponse::ok()
+                                                .content_type("text/html")
+                                                .unwrap()
+                                                .body(&crate::support::consent_page_html(
+                                                    "/authorize".into(),
+                                                    pre_grant,
+                                                )),
+                                        )
+                                    },
+                                )))))
```
HeroicKatora: Instead of sending the compete box here, we could also send a simple enum variant.

```
enum OwnerInteraction {
    NonInteractiveRequest,
    Post(String),
}
```

Then the `State` can construct the correct boxed solicitor in its `Handler` implementation itself, which would fit its role better. This also removes the need to typedefing the `Extras`.

HeroicKatora: The `GET` handler here could also have a comment how that request could be replayed. In particular, only `POST` is Cors protected by default. I think that would be instructive.

oxide-auth-actix/examples/actix-example/src/main.rs
```diff
@@ -0,0 +1,236 @@
...
+            registrar: &mut self.registrar,
+            issuer: &mut self.issuer,
+            solicitor,
+            scopes: self.scopes.clone(),
+            response: OAuthResponse::ok,
+        })
+    }
+}
+
+impl Endpoint<OAuthRequest> for State {
```
HeroicKatora: If the other code also uses `Generic`, *maybe* switch to it for other code flow as well.

What are the pros and cons of having an implementation of `Endpoint` in the example?

oxide-auth/src/frontends/simple/endpoint.rs
```diff
@@ -114,6 +116,17 @@ pub struct Generic<R, A, I, S=Vacant, C=Vacant, L=Vacant> {
     pub response: L,
 }
 
+/// A simple wrapper around an Endpoint to change it's error type into anything `Into`-able.
+pub struct EndpointErrorMapper<E, Error>(E, PhantomData<Error>);
```
HeroicKatora: Name bikeshedding: `MapErr`

It is terser and already in use for `futures`. `Map` is utilized in a number of places in the standard library as well and would also count as reasonably similar. The module path also contains `endpoint` already.

oxide-auth-actix/examples/actix-example/src/main.rs
HeroicKatora: Oh right. The enum could contain three variants so as to avoid the need of wrapping it into a `Option`.

oxide-auth/src/frontends/simple/endpoint.rs
asonix: Would it make sense, then, to add a `.map_err` method to the `Endpoint` trait?

HeroicKatora: I've thought about adding it to a separate `EndpointExt` trait, yes. I just realized that this is not quite as generic as `MapErr` so maybe we shouldn't change this right now.

There can always be a typedef for this to refer to a more generic replacement, later.

oxide-auth-actix/examples/actix-example/src/main.rs
asonix: I like not needing to name all the types on Generic all the time. I supposed I could just `Box<dyn Endpoiint<OAuthRequest, Error = WebError>>`, though

oxide-auth/src/frontends/simple/endpoint.rs
asonix: Actually, it might be more similar to `.from_err()`

HeroicKatora: Remove the `Endpoint` portion of the name, however.

HeroicKatora: `ErrorInto` would also be fine.

oxide-auth-actix/examples/actix-example/src/main.rs
HeroicKatora: You can utilize [`Vacant`](https://docs.rs/oxide-auth/0.4.5/oxide_auth/frontends/simple/endpoint/struct.Vacant.html) to return a `Generic` without a solicitor but still annotate the return type as 
> `-> impl Endpoint<OAuthRequest, Error=WebError>` 

```
Generic {
…
    solicitor: Vacant,
}
```

HeroicKatora: Well, technically it then has a solicitor that denies everything but that should not make a functional difference where it is used.

asonix:
I could definitely see a use for more types like `ErrorInto` on an `EndpointExt` trait.

```
WithSolicitor<E, S>(E, S)

EndpointExt::with_solicitor(endpoiont, new_solicitor)
```

Maybe that's a discussion for another day

oxide-auth-actix/examples/actix-example/src/main.rs
asonix: What sort of comment should be there? just like `// Watch out! Requests for this resource are repeatable`

HeroicKatora: Unfortunately the `post` is of course not protected perfectly against CSRF as well but better. So something along the lines of:

> GET requests should not mutate server state and are extremely vulnerable accidental repetition as well as Cross-Site Request Forgery (CSRF).

PR #53, opened 2019-09-03
Next ring versions

HeroicKatora:
This fixes `oxide-auth` being incompatible with `ring = 0.14`. Note that this is a bit hacky and there should be a better solution when/if the version disparity gets even greater in the future. In the meantime most web frameworks use the same `cookies` crate which only has `0.13` and `0.14` in relevant versions and is reasonably stable.

Removes the pulic dependency on `ring` to allow these changes in the future.

- [x] This change has tests

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -33,6 +33,12 @@ nightly_task:
   test_script: cargo test -p "$CRATE"
   before_cache_script: rm -rf $CARGO_HOME/registry/index
 
+ring_task:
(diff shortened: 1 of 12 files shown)


HeroicKatora:
cc: @asonix

asonix:
I am impressed by this trick to get 0.13 and 0.14 to both compile tbh! looks good to me, and also works in my own projects via git depends

PR #56, opened 2019-09-29
Replace ring::Unspecified with RegistrarError

### LINKED ISSUE #54: Don't expose Ring types in public API
# Project Improvement
Currently, the `PasswordPolicy` trait requires the use of ring's `Unspecified` type, which means any implementors, regardless of dependency on Ring, need to import Ring to return that type.

This isn't a huge issue, since oxide-auth depends on ring already and this isn't adding dependency bloat per-se, but `()` for an error type makes about as much sense as `Unspecified`, and there's precedent for returning `()` errors in other parts of the codebase.

HeroicKatora:
There is one instance where Unspecified was replaced with a
PrimitiveError. On checking a password the hash is checked for internal
consistency and a failure to validate indicates that the policy was
changed during operation. This is however not the failure of the user
but should be an internal server error.

This fixes #54 .

### DIFF
--- a/oxide-auth/src/primitives/registrar.rs
+++ b/oxide-auth/src/primitives/registrar.rs
@@ -181,7 +181,7 @@ impl fmt::Debug for ClientType {
     }
 }
 
-impl From<Unspecified> for RegistrarError {
+impl RegistrarError {
     fn from(err: Unspecified) -> Self {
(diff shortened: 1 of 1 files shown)


HeroicKatora:
cc @asonix

asonix:
I didn't see this until just now, oops! let me see if I can reconcile this with my other PR

Issue #55, opened 2019-09-24
Ring question


asonix:
How would you feel about a PR replacing `ring` with [`sha2`](https://docs.rs/sha2/0.8.0/sha2/), [`hmac`](https://docs.rs/hmac/0.7.1/hmac/), and [`rand`](https://docs.rs/rand/0.7.2/rand/), and moving the default `pbkdf2` type into a separate crate that depends on ring

HeroicKatora:
Interested, yes. But there are some concerns around the details of the details of their replacement. I certainly want to remove `ring` from the public interface and then maybe make it into an optional feature instead of replacing it with another implementation.

HeroicKatora:
To be more precise, it's not clear to me to which extent pure Rust implementations of signing with `hmac` and `sha2` are side channel resistant while `ring` takes some care in this regard. I'm a little more comfortable with replacing `pbkdf2` as the default password storage function.

gzp-crey:
As actix (ex session handling) is using ring 0.16, is there any plan how to manage all the ring versions ?
I can update it for actix, but it'll break all the other frameworks.

HeroicKatora:
I'm open for PRs that gradually move away from `ring` as the supplier of crypgraphic operations. The signature operations with hmac would be a good start. My personal impression was that `ring` takes security more seriously than the other crates. (E.g. given that Rust's story for constant time is more based on guesswork than compiler guarantees, real assembly is the most secure way to perform cryptography right now).

PR #57, opened 2019-09-29
Split ring into separate crate

asonix:
This change splits ring, and types based on ring-backed types, into another crate to avoid a hard dependency on Ring. A lot of the concrete implementation does depend on ring, so a good deal of tests have moved over into the ring crate, also

 - [x] Corresponds to issue #55 


HeroicKatora:
I've been a bit conflicted regarding these changes. After a bit of consideration I think this is missing the tree for the forest. Since this moves *all* provided implementations of some traits to the `-ring` trait it doesn't actually help avoid `ring` in the depency tree unless you were to only use custom implementations anyways. But then a simple feature switch to enable the `ring` implementations would be potent enough?

There are a number of points I'd want to see addressed individually:
* Some changes to the interface are also done here. What is the reasoning behind them? If they are necessary then they should be done regardless of the split and separately from it.
* Tests are also moved (since no implementations of some primitives exist). This *drastically* reduces coverage.
* One positive thing, I'm much more open-minded towards replacing `ring` with a combination of other crypto library if the `ring` implementations are kept available somewhere else. Purely adding the `-ring` crate for now also makes it easier to review.
* The `-ring` crate has no problem comitting to a `ring` depency in its interface. That means some utility constructors from `hmac::Key` that were avoided could be added back?

asonix:
I do have an additional branch which implements the extent of the `-ring` crate with rustcrypto libs. It make additional changes to the API, specifically requiring that `registrar` be mutable.

I think my reasoning for changing the API in this PR was to prepare a bit for that, but you're right that this PR specifically doesn't (and shouldn't) depend on the API change.

asonix:
And also, doesn't it make sense that the tests would test the implementation, rather than the interface? In a statically-typed language like Rust, the interface is well-defined just by existing, is it not?

I'd also argue that test coverage has not reduced, since the same code is being tested with the same tests, they're just somewhere else in the file structure

asonix:
Here's the actual changes for adding rustcrypto https://github.com/Aardwolf-Social/oxide-auth/pull/1

it's likely that things could be done _better_, since some types are duplicated across -ring and -rustcrypto that could be defined in the base crate with a generic

HeroicKatora:
> It makes additional changes to the API, specifically requiring that registrar be mutable.

Yes I noted but why?

HeroicKatora:
To be more precise, what are you trying to achieve with the API change? Why should `check` have to mutate the data structure for example?

HeroicKatora:
> And also, doesn't it make sense that the tests would test the implementation, rather than the interface? In a statically-typed language like Rust, the interface is well-defined just by existing, is it not?

> I'd also argue that test coverage has not reduced, since the same code is being tested with the same tests, they're just somewhere else in the file structure.

They are in another crate and the new structure and removal of the default features make it impossible to test the interface of `oxide-auth` alone. I find that extremely irritating. In that case every addition or change in `oxide-auth` needs to be immediately adapted for in the `-ring` crate to allow it to be tested. That's creates a hard dependency that (I think) shouldn't exist between these separate crates even though they share the same workspace. And release are no longer self-contained. Usually I can download a crate's published files to also run its testsuite.

HeroicKatora:
Note: (Integration) tests use the `[dev-dependencies]` so with a dependency cycle this may work without moving the tests themselves. At least that makes the review of the changes easier. I'm still not sure how well the workspace can cope with `[dev-dependencies]` requiring a version different from the one that exists locally (on the offchance of this being required for some reason)?

asonix:
The reason for making some of these methods mutable is due to the RustCrypto and Rand implementations' reliance on having mutability. For example, things implementing rand's `Rng` trait (e.g. StdRng) need to be mutable to use `try_fill` when filling a buffer. Things implementing RustCrypto's Digest trait (e.g. Sha256) need to be mutable to hash and verify.

asonix:
Not _all_ implementations of these traits need mutability (ring didn't, obviously), but some implementations might.

HeroicKatora:
> The reason for making some of these methods mutable is due to the RustCrypto and Rand implementations' reliance on having mutability. For example, things implementing rand's Rng trait (e.g. StdRng) need to be mutable to use try_fill when filling a buffer. Things implementing RustCrypto's Digest trait (e.g. Sha256) need to be mutable to hash and verify.

For the case of `Rng` use *internal* mutability. That's precisely what it was made for. Most rng implementations will still not require it because they are in thread local storage and so fully `Sync` or rather they have their own mutability. I don't get how `Digest` influence this at all since it should just be on the stack during the evaluation of `check` but not have any other state.

asonix:
In the case of the RustCrypto implementation of `Assertion`, it carries around a `Hmac<Sha256>` internally, which must be mutable to sign data.
```rust
self.secret.input(&to_sign);
self.secret.result_reset()
```
This could be avoided by just using `Hmac::<Sha256>::new_varkey(key_bytes)` each time `signature` is called, but that seems like not a great idea (the ring implementation carries the SigningKey type around).

asonix:
Now that I look at it again, I'm not convinced PasswordPolicy's `store` method needs to be mutable, which would also mean `check` would not need to be mutable. Maybe an earlier iteration of the code needed this, but that is no longer true.

asonix:
I think I was hesitant to clone the Hmac because I've been burned by cloning StdRng in the past, but you're right that Hmac itself is likely safe to clone, since it's deterministic and doesn't contain randomness.

asonix:
I think what I will do this evening is I will re-create this PR without changes to the interface, and maybe a circular dev-dependency to leave more of the tests in the primary crate.

PR #58, opened 2019-10-02
Split Ring 2

asonix:
A redo of #57 

### What's different?
- Less has moved
- TokenSigner is now generic over T with the help of new traits.
- Tests have been put in the library-level tests directory so they can properly access oxide_auth_ring
- no existing traits have changed signatures

### checklist
 - [ ] This change has documentation (a couple things need documenting, still)
 - [x] Corresponds to issue (number)


asonix:
@HeroicKatora what do you think of this version

HeroicKatora:
1. It's still a pity to require an external crate for basic usage of `Registrar`, `Issuer`, `Authorizer`. The main concern that this doesn't avoid depending on `ring` still stands, it only shifts the dependency to yet another crate that needs to be dependend on anyways? I really need some stronger a better motivation and explanation on how it is helpful without having some more compatible implementation in the base crate.

2. If there is no default `PasswordPolicy` then it should become a constructor parameter. That avoids having an error type on the other methods as well. It's currently an `Option<Box<_>>` only because that conveniently avoids an allocation in the constructor.

3. `oxide-auth-ring` should use the interface before the public `ring` dependencies were removed. It has no interest in hiding that dependency. In particular, its `Assertion` may want to just accept an `ring::hmac::Key`.

4. `SerdeAssertionGrant` exists as an implementation detail, it shouldn't need to be public. If it is used as a public encoding mechanism, it should just be a method `fn encode(&Grant) -> Vec<u8>`.

5. `Signer`/`Tagged` seems overengineered. I think the cleaner interface would simply be a `&str` parameter on `Signer` as [authenticated, associated data](https://en.wikipedia.org/wiki/Authenticated_encryption) and only use the `TaggedAssertion` as an implementation detail of this.

oxide-auth/src/primitives/generator.rs
```diff
@@ -44,180 +36,68 @@ pub trait TagGrant {
...
 
-/// Generates tokens from random bytes.
-///
-/// Each byte is chosen randomly from the basic `rand::thread_rng`. This generator will always
-/// succeed.
-pub struct RandomGenerator {
-    random: SystemRandom,
-    len: usize
+/// An implementation of TagGrant backed by a cryptographically secure source of randomness
+pub struct RandGenerator<T> {
```
HeroicKatora: Having `rand` implementations makes the basic interfaces more useful. The names don't need to be differ and I prefer having a single, known good implementation to overly configurable systems. What is gained from the parameter? The instantiation should be a perfectly unpredictable generator by design, all of which should be indistinguishable if well seeded. This only adds a point of failure. To quote from [`rand::CryptoRng`](https://docs.rs/rand/0.7.2/rand/trait.CryptoRng.html) which acknowledges this issue:

> Note that this trait is provided for guidance only and cannot guarantee suitability for cryptographic applications. In general it should only be implemented for well-reviewed code implementing well-regarded algorithms.

This seems to suggest that it is *unsafe* to allow generic instantiations.

HeroicKatora: Note: A renaming to `WeakRandGenerator` may be fine. At least it makes the potential consequences and the level of scrutiny applied to the implementation clear from the start.

asonix: What if instead of making it generic over CryptoRng I put a Box<dyn Rng> in it and only exposed methods for creating with StdRng (Send+Sync) or ThreadRng (!Send+!Sync)

asonix: Maybe with a method like `fn unsafe_custom_rng<R: CryptoRng + Rng>(&mut self, rng: R)`

asonix: I guess boxing wouldn't work with the sync amd send bounds, though

asonix: Just going with StdRng might be the best option, since it's Send+Sync, part of the Rand crate, and officially marked CryptoRng

HeroicKatora: That were my thoughts as well if there is no obvious advantage of `ThreadRng`.

asonix:
I'm going to close this. I don't think splitting ring out brings enough benefit right now, but if the disparity in ring dependencies increases in the rust ecosystem in the future, it can be revisited

PR #62, opened 2019-12-10
Port to acitx_web 2.0-alpha

gzp-crey:
As actix and actor model is not supported by actix_web at the moment and alternative solution is provided using future mpsc, oneshot channels.

This changes/fixes/improves #37;

State is a "shared" resources and the actor send is emulated by the mpsc channel.
The response from the State is returned to the endpoint through a oneshot channel.


HeroicKatora:
I don't follow the development of `actix` closely, how likely is a breaking change from the current pre-release version until the release of `2.0`?

asonix:
I think the actix-web portions are probably pretty close to final. The biggest pain is going to be Ring 0.16, though

gzp-crey:
For ring the changes are quite trivial, a few renames only. But it won't be compatible to the other web frameworks at the moment. Also I have no idea how to handle this without introducing a feature to select ring version.

gzp-crey:
a new port is created

PR #63, opened 2019-12-13
Ring 0.16 and actix 2.0-alpha support

gzp-crey:
This is not a full PR and it shall not be merged into master. 
Related to  #37.
This PR is only for others who wants to use this crate in combination with actix_web 2.0(alpha) and actix-session.

Also note that actix-session was add only to check crate dependencies and not used by any of the crates in this repo.


gzp-crey:
closing as it is outdated

Issue #71, opened 2019-12-23
Update to actix-web 2.0


HeroicKatora:
# Project Improvement

actix-web 2.x has some breaking changes and comes with support for standard futures.

## Other context

See prior comments in #37. It also requires an update to `ring` to support `v0.16`.

# Tracking pull request

 - [x] A pull request is #72
 - Related: #63 
- [x] Motivates: #70

gzp-crey:
The error handling part (ex Resource grant) looks awkward a little bit:
```
 {
        Ok(Ok(_grant)) => ...  // granted
        Ok(Err(Ok(e))) => ... // denied
        Ok(Err(Err(e))) => ... // "internal" error during grant (ResourceFlow)
        Err(e) => ... // "internal" error in handler ex. mailbox error
    }
```

gzp-crey:
Is there anything that block the related PR to be merged back to main? (#72)

thespooler:
Since #37 is closed and dicussion moved here, I'm also wondering about that.
The actix-web 2.0 update has been merged a few minutes ago (thanks @HeroicKatora).
Without bumping to ring 0.16.9, actix-identity and actix-session can't be used as they force actix-web/actix-http with feature session-cookie, wich requires 0.16.9.

Is there any path forward?

HeroicKatora:
Yes, bite the bullet and replace `ring` with [`RustCrypto/*`](https://github.com/RustCrypto/) in the main crate, and create a `oxide-auth-ring` side crate with the `ring`-based code (which can then probably also be updated to a much newer version of `ring`).

HeroicKatora:
The question is whether it really is sensible (from a security standpoint) to provide _all_ existing implementations of authorizer and issuer or to drop some from the main crate. The map-based implementations are fine without having to rely on any the security of any crypto implementation so one could say that the encrypting ones are only supplementary and could live in another optional crate. The registrar implementation is a slight concern but really, pbkdf2 is not exactly state-of-the-art in any case. Maybe replace it with blake3 given that there is an **official** [Rust implementation](https://github.com/BLAKE3-team/BLAKE3).

thespooler:
Quite the undertaking!
Why the side crate, is there really a hard dependency on someting from ring explicitly that doesn't have a RustCrypto equivalent? If so, could it really be decoupled from the main crate without introducing the same ring dependency, but one crate further? In other words, is it worth it? If we have an alternative, is keeping a ring impl something desirable?

[...]

PR #72, opened 2019-12-28
bump to actix-web 2.0

gzp-crey:
This changes/fixes/improves #37, #71;

Actor model works for actix again. 
This PR tries to minimize the change and use the old model.


HeroicKatora:
Sorry for the delay, it was partly me not realizing one of them had *no* bump to `ring`. New version of `oxide-auth-actix` will probably be published later today.

PR #73, opened 2019-12-30
Ring 0.16 actix web 2.0

gzp-crey:
An incomplete PR to bump vesrions for ring to 0.16 and actix_web 2.0
This is related to #37, #71, #55

Only actix backend is supported at the moment.


gzp-crey:
triage - outdated for me

PR #77, opened 2020-02-20
Set the visibility of OAuthResponse fields to pub

hioki:

I changed the visibility of OAuthResponse fields to `pub` because there was no way to access `OAuthResponse` response header or response body from outside.

By the way, I want to do the following things.

1. At the time of `POST /authorize`, keep the information that linked the issued authorization code and the information of the authenticated user as HashMap.
2. At the time of subsequent `POST /token`, the user at the time of authentication is specified from the authorization code specified in the request body using the HashMap described above, and the issued access token is associated with the user at the time of authentication as another HashMap.
3. After that, acquire the user information at the time of authentication from the access token at the time of request as needed and use it.

### DIFF
--- a/oxide-auth-actix/src/lib.rs
+++ b/oxide-auth-actix/src/lib.rs
@@ -119,6 +119,18 @@ pub struct OAuthRequest {
     body: Option<NormalizedParameter>,
 }
 
+impl OAuthResponse {
+    /// Get the headers from `OAuthResponse`
+    pub fn get_headers(&self) -> HeaderMap {
+        self.headers.clone()
+    }
+
+    /// Get the body from `OAuthResponse`
+    pub fn get_body(&self) -> Option<String> {
+        self.body.clone()
+    }
+}
+
 /// Type implementing `WebRequest` as well as `FromRequest` for use in guarding resources
 ///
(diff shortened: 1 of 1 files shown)


hioki:
@HeroicKatora Thanks for merging #78 and #79 ! I hope you see this one too!

HeroicKatora:
Yes, I've seen it but don't have the time to evaluate it. In general, I have some reservations towards making all fields public for two reasons:

* It commits to this _exact_ set of fields and allows it to be trivially constructed. This could be solved by keeping one private field or marking the struct `#[non_exhaustive]` but this solutions might muddle the interface and ultimately hurt usability.
* It's prevents having internal invariants under. Furthermore, the use case makes a case for _reading_ these attributes but it doesn't seem clear if it argues for being able to mutate the attributes as well.

These two might not be applicable to the struct in question. Maybe a trivial public constructor is desirable. Maybe no internal invariants are needed. I'll try to get to evaluating this soon.

hioki:
Oooh...sorry, I casually specified `pub`.

I'm sorry that you've pointed out before to be careful with the public interface at https://github.com/HeroicKatora/oxide-auth/pull/79#discussion_r383128191.

In my use case, I actually don't need `status`, it just needs to get the `headers` and `body`. Therefore, I'm going to define each getter.

PR #79, opened 2020-02-24
Multiple redirect uris in client

hioki:

Added support for registering multiple Redirect URIs.

### DIFF
--- a/oxide-auth/src/primitives/registrar.rs
+++ b/oxide-auth/src/primitives/registrar.rs
@@ -121,7 +121,7 @@ pub enum RegistrarError {
 #[derive(Clone, Debug)]
 pub struct Client {
     client_id: String,
-    redirect_uri: Url,
+    redirect_uri: Vec<Url>,
     default_scope: Scope,
     client_type: ClientType,
 }
@@ -137,7 +137,7 @@ pub struct EncodedClient {
     pub client_id: String,
 
     /// The registered redirect uri.
-    pub redirect_uri: Url,
+    pub redirect_uri: Vec<Url>,
 
     /// The scope the client gets if none was given.
     pub default_scope: Scope,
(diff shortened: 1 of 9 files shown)


oxide-auth/src/primitives/registrar.rs
```diff
@@ -189,15 +189,15 @@ impl RegistrarError {
 
 impl Client {
     /// Create a public client.
-    pub fn public(client_id: &str, redirect_uri: Url, default_scope: Scope) -> Client {
```
HeroicKatora: I think this unecessarily breaks the public interface. Also, I do have the concern that clients must have *at least one* redirect `Url`, so I'd rather see instead a added method to provide additional `Url`s. This also makes it clear which is the default uri that gets substituted when the parameter is missing.

```rust
impl Client {
    fn with_alternative_uris(self, uris: Vec<Url>) -> Self;
}
```

oxide-auth/src/primitives/registrar.rs
```diff
@@ -137,7 +137,7 @@ pub struct EncodedClient {
     pub client_id: String,
 
     /// The registered redirect uri.
-    pub redirect_uri: Url,
```
HeroicKatora: I'll have to think if this is the best way to approach this, possibly better to add a new attribute.

oxide-auth/src/primitives/registrar.rs
```diff
@@ -522,16 +522,15 @@ impl Registrar for ClientMap {
         };
 
         // Perform exact matching as motivated in the rfc
-        match bound.redirect_uri {
-            None => (),
-            Some(ref url) if url.as_ref().as_str() == client.redirect_uri.as_str() => (),
+        let redirect_uri = match bound.redirect_uri {
+            None if client.redirect_uris.len() == 1 => Cow::Owned(client.redirect_uris.first().unwrap().clone()),
```
HeroicKatora: `first().unwrap()` is easier to read as `[0]` in any case.

PR #81, opened 2020-04-09
Replace pbkdf2 by Argon2 in registrar

thespooler:
This change replaces the current PBKDF2 password by Argon2 from the argonautica crate.

 - [ ] This change has tests 
 - [x] Corresponds to issue #71 with discussions in #55

The tests looks good, so I did not find anything to add.
The only thing that I recon might be a point of concern is that store and check handles `[u8]`, while `Hasher` operates on `String`, so I handle the utf8 conversion. 
I'm not quite sure that no security implications are hidden in there.


oxide-auth/Cargo.toml
```diff
@@ -14,8 +14,10 @@ license = "MIT OR Apache-2.0"
 autoexamples = false
 
 [dependencies]
+argonautica = "0.2.0"
 base64 = "0.11"
 chrono = "0.4.2"
+lazy_static = "1.4.0"
```
HeroicKatora, line 20: I'd prefer [`once_cell`](https://crates.io/crates/once_cell) here. It's the same effect without macro. The compiler is also currently switching to it in their implementation.

oxide-auth/src/primitives/registrar.rs
```diff
@@ -291,107 +286,69 @@ pub trait PasswordPolicy: Send + Sync {
...
 
-impl Clone for Pbkdf2 {
-    fn clone(&self) -> Self {
-        Pbkdf2 {
-            random: Some(SystemRandom::new()),
-            .. *self
+impl<'a> Default for Argon2<'a> {
+    fn default() -> Self {   
+        Argon2 {
+            secret_key: "Please provide a SecretKey! This is for your own protection.".into()
```
HeroicKatora, line 298: This doesn't really add any protection. This pepper, in terms of password storage, is useful to have but it should either

* Cause an actual error if it is unset.
* Allow being used without a secret key. This is _basically_ what the current function is doing so it won't provide any worse security than at the moment. Would an ephemeral implementation be useful, even though credentials should be long lived?.

oxide-auth/src/primitives/registrar.rs
```diff
@@ -291,107 +286,69 @@ pub trait PasswordPolicy: Send + Sync {
...
-    fn default() -> Self {
-        Pbkdf2 {
-            random: Some(SystemRandom::new()),
-            .. *PBKDF2_DEFAULTS
-        }
-    }
+/// Store passwords using `Argon2` to derive the stored value.
+#[derive(Debug)]
+pub struct Argon2<'a> {
+    secret_key: SecretKey<'a>
```
HeroicKatora, line 292: I'd rather avoid the lifetime parameter and specify the internal one as `'static`. Note that there exists a method [`to_owned`](https://docs.rs/argonautica/0.2.0/argonautica/input/struct.SecretKey.html#method.to_owned) so this can be done in any case.

oxide-auth/src/primitives/registrar.rs
```diff
@@ -291,107 +286,69 @@ pub trait PasswordPolicy: Send + Sync {
...
     }
 }
 
-impl fmt::Debug for Pbkdf2 {
-    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        f.debug_struct("Pbkdf2")
-            .field("iterations", &self.iterations)
-            .field("random", &())
-            .finish()
+impl From<std::string::FromUtf8Error> for RegistrarError {
```
HeroicKatora, line 303: Those impls are part of the crate's interface. I find this one a bit confusing (although it might be useful in general).

thespooler:
Argonautica seems to require clang, wich is missing from CI.
I don't know how to fix that tho.

oxide-auth/src/primitives/registrar.rs
```diff
@@ -291,107 +286,69 @@ pub trait PasswordPolicy: Send + Sync {
...
-
-        match self.random.as_ref() {
-            Some(random) => random.fill(&mut rnd_salt),
-            None => SystemRandom::new().fill(&mut rnd_salt),
-        }.expect("Failed to property initialize password storage salt");
-
-        vec.extend_from_slice(user_identifier);
-        vec.extend_from_slice(&rnd_salt[..]);
-        vec
+impl From<argonautica::Error> for RegistrarError {
```
HeroicKatora, line 309: This impl leaks the implementation detail of `argonautica` into the public interface. That should be avoided.

oxide-auth/src/primitives/registrar.rs
thespooler: It is opt-out and would fail if no secret is provided. 
So users should be **required** to provide one?

```
        .opt_out_of_secret_key(true); // Default is `false`
        // 👆 As an extra security measure, if you want to hash without a secret key, which
        // is not recommended, you must explicitly declare that this is your intention
        // by calling this method and setting the `opt_out_of_secret_key` configuration to
        // `true` (by default, it is set to `false`); otherwise hashing will return an error
        // when you fail to provide a secret key
```

HeroicKatora: The crate also documents that the underlying C implementation does not use a secret key for the purpose of password hashing, if I read it correctly. I'd take that as meaning it is fine to opt-out of it.

oxide-auth/Cargo.toml
thespooler: Nice, didn't know that one.

oxide-auth/src/primitives/registrar.rs
thespooler: I did miss to_owned.

oxide-auth/src/primitives/registrar.rs
thespooler: Would Option<SecretKey<'static>> sound good to you?
That way we can trap Some(key) and use it and otherwise opt_out_of_secret_key.

oxide-auth/src/primitives/registrar.rs
thespooler: I will remove both impl

oxide-auth/src/primitives/registrar.rs
thespooler: Removed

oxide-auth/src/primitives/registrar.rs
HeroicKatora: `'static` as a lifetime does not mean that there is only one. It means that the secret key is _owned_.

oxide-auth/src/primitives/registrar.rs
HeroicKatora: That sounds good.

oxide-auth/src/primitives/registrar.rs
thespooler: Are you sure? If so, my bad then, but the [rustbook](https://doc.rust-lang.org/1.9.0/book/lifetimes.html#static) says this:
> The lifetime named ‘static’ is a special lifetime. It signals that something has the lifetime of the entire program. [...] String literals have the type &'static str because the reference is always alive: they are baked into the data segment of the final binary.

Or am I misunderstanding?

HeroicKatora: Yes, you are. In two ways: It's true that a _reference_ with lifetime `'static` refers to data that lives for the entire program. But if something owns the data, e.g. a `Vec<u8>` does, then it is totally under that types control when the data's lifetime ends and it can decide that it lives for the whole program. This _also_ has by definition a `'static` lifetime. This works:

```rust
fn assert_static<T: 'static>() {}
assert_static::<Vec<u8>>();
```

Secondly, this has nothing to do with uniqueness in any case.

thespooler: Ok, it is still not totally clear to me, but I can see it makes sense.
Last commit removed `'a` and replaced it with `'static` on `SecretKey`

oxide-auth/src/primitives/registrar.rs
thespooler: Fixed

oxide-auth/Cargo.toml
thespooler: Fixed

HeroicKatora:
> Argonautica seems to require clang, wich is missing from CI.
I don't know how to fix that tho.

It should be possible to `apt install` the necessary packages. I'm not quite sure why `clang` in particular is required.

thespooler:
It seems to work with the reference implementation of phc-winner-argon2

thespooler:
Ok, I think I got this. Except the two tasks that were `Signaled to exit!`. 
I guess some kind of resource limiting on the CI host.

thespooler:
This PR is obsolete with #85 merged and #86.

Issue #83, opened 2020-04-10
Replace ring by rust-argon2, rand, hmac and sha2


thespooler:
# Feature

As discussed in issue #55 and #71, it seems that moving away from ring is the most sensible way to integrate with other projects that might have conflicting package version requirements.

## Alternatives

Leave ring, split crates, try to follow other projects dependencies and match ring versions accordingly.

## Context

actix-web 2.0 when used with actix-identity or actix-session requires ring 0.16.9, while oxide-auth requires ring < 0.15.

## Tracking pull request

 - [x] Replace pbkdf2 by Argon2 in registrar #86
 - [x] Replace ring in generator #85
 - [x] Replace ring in registrar #86 
 - [x] Replace ring in pkce code_grant extension #85
 - [x] Remove ring from dependencies and tests #87 

### Related 
 - [ ] Update to actix-web 2.0 #71
 - [ ] Ring question #55
 - [ ] Ring 0.16 actix web 2.0 #73

PR #84, opened 2020-04-12
Replace ring by argonautica, rand, hmac and sha2

thespooler:
This replaces all usages of ring by their RustCrypto equivalent, namely rand, hmac and sha2.
This PR also includes #81, replacing pbkdf2 by argonautica, so as to remove ring from Cargo.toml and libs.rs, completely removing any ring dependency.

 - [x] Corresponds to issue #83 


thespooler:
Take a look at `primitives/generator.rs`, since there is no equivalent to `SecretKey` in `hmac`, I did couple the key and hashing selection mechanism in a few places (`new`, `ephemeral`, `extract` and `signature`). Meaning that to add another option beside `Hmac<Sha256>` isn't as easy as the signature of `Assertion::new(AssertionKind, key)` might lead you to believe.
If you have any idea on how to neatly wrap this, let me know.

HeroicKatora:
Would you mind making the changes independent of the `bpkdf2` removal? I still haven't made up my mind on whether introducing a native dependency (libclang) is good, and want to find out why it is even necessary in the used library. Yes, this means that `ring` won't be fully removed by this PR but that's fine. Better to go slow and steady than rush.

oxide-auth/src/code_grant/extensions/pkce.rs
```diff
@@ -166,13 +165,12 @@ impl Method {
...
+               if encoded.as_bytes() == verifier.as_bytes() { Ok(()) } else { Err(()) },
             Method::Sha256(encoded) => {
-                let digest = digest(&SHA256, verifier.as_bytes());
-                let b64digest = b64encode(digest.as_ref());
-                verify_slices_are_equal(encoded.as_bytes(), b64digest.as_bytes())
-                    .map_err(|_| ())
+                let mut hasher = Sha256::new();
+                hasher.input(verifier.as_bytes());
+                let b64digest = b64encode(&hasher.result());
+                if encoded.as_bytes() == b64digest.as_bytes() { Ok(()) } else { Err(()) }
```
HeroicKatora, line 173: Don't remove the constant time verification here.

oxide-auth/src/primitives/generator.rs
```diff
@@ -49,22 +49,22 @@ pub trait TagGrant {
...
     pub fn new(length: usize) -> RandomGenerator {
         RandomGenerator {
-            random: SystemRandom::new(),
+            random: OsRng {},
             len: length
         }
     }
 
-    fn generate(&self) -> String {
+    fn generate(&mut self) -> String {
```
HeroicKatora, line 65: That's a surprising change in signature and leads to a bunch of unwarranted removals of impls down the road. `OsRng` is `Copy` and there should be no need for that.

oxide-auth/src/primitives/generator.rs
```diff
@@ -132,24 +132,24 @@ impl Assertion {
...
+    /// If future suites are added where this is not possible, this function may panic when supplied
     /// with an incorrect key length.
+    /// 
+    /// Currently, the implementation lacks the ability to really make use of another hasing mechanism than
+    /// hmac + sha256.
     pub fn new(kind: AssertionKind, key: &[u8]) -> Self {
-        let key = match kind {
-            AssertionKind::HmacSha256 => SigningKey::new(&SHA256, key),
+        match kind {
+            AssertionKind::HmacSha256 => Assertion { key: key.to_vec() },
```
HeroicKatora, line 142: `SigningKey::new` did perform some essential verification and transformation depending on the length, where has this gone? Read [the documentation of `ring`](https://docs.rs/ring/0.16.12/ring/hmac/struct.Key.html#method.new) on what it does and which standard this implements.

oxide-auth/src/primitives/generator.rs
```diff
@@ -132,24 +132,24 @@ impl Assertion {
...
-            secret: key,
         }
     }
 
     /// Construct an assertion instance whose tokens are only valid for the program execution.
     pub fn ephemeral() -> Self {
-        Assertion {
-            secret: SigningKey::generate(&SHA256, &SystemRandom::new()).unwrap(),
-        }
+        // XXX Why 64?
```
HeroicKatora, line 149: There should be a definition on the key size required by hmac-sha256. This will explain the size, I'm currently confused as to why this is not `32`.

oxide-auth/src/primitives/generator.rs
```diff
@@ -160,14 +160,21 @@ impl Assertion {
     fn extract<'a>(&self, token: &'a str) -> Result<(Grant, String), ()> {
         let decoded = decode(token).map_err(|_| ())?;
         let assertion: AssertGrant = rmp_serde::from_slice(&decoded).map_err(|_| ())?;
-        ring::hmac::verify_with_own_key(&self.secret, &assertion.0, &assertion.1).map_err(|_| ())?;
+
+        let mut hmac = Hmac::<sha2::Sha256>::new_varkey(&self.key).unwrap();
```
HeroicKatora, line 164: Maybe this performs the key-transformation? If so, store this struct and not the raw key. If it is possible to clone it here, that is.

oxide-auth/src/primitives/generator.rs
```diff
@@ -232,21 +241,30 @@ impl TagGrant for RandomGenerator {
...
+//     }
+// }
 
 impl TagGrant for Rc<RandomGenerator> {
     fn tag(&mut self, _: u64, _: &Grant) -> Result<String, ()> {
-        Ok(self.generate())
+        if let Some(rng) = Rc::get_mut(self) {
+            Ok(rng.generate())
+        } else {
+            Err(())
```
HeroicKatora, line 256: Why remove on impl but let the other error? Using `Rc::get_mut` is just brittle.

thespooler:
About argonautica, there's also argon2rs and rust-argon2. From what I see, argon2rs seems decent, pure rust and benchmarked.

oxide-auth/src/primitives/generator.rs
thespooler: https://github.com/RustCrypto/MACs/blob/master/hmac/src/lib.rs#L142
```
        // The key that Hmac processes must be the same as the block size of the
        // underlying Digest. If the provided key is smaller than that, we just
        // pad it with zeros. If its larger, we hash it and then pad it with
        // zeros.
```
Hmac seems to pad or truncate (via hashing) the key on its own. I did not think it was necessary to actually do it with the key. Am I missing something?

oxide-auth/src/primitives/generator.rs
thespooler: Oh yeah, cause I'm dumb. I was intending to find how to get to `Mac::KeySize` for a given impl (say `Hmac::<Sha256>`), but I'm not proficient enough in rust to figure it out and I forgot to get back to it.

oxide-auth/src/primitives/generator.rs
thespooler: That might be the right thing to do, but I figured that since you used `SecretKey` and that wrapped the intent to use SHA256 at construction time and I couldn't figure out how to get a similar result, that having the key available for reuse might be of some use. It does feel wrong tho. That's what I alluded to in my 1st comment.

oxide-auth/src/primitives/generator.rs
thespooler: Totally right, will `Clone()`.

oxide-auth/src/primitives/generator.rs
thespooler: You were right with the `&self mut` on `generate`. I reverted and fixed this.

oxide-auth/src/code_grant/extensions/pkce.rs
thespooler: I'll try with `subtle` `ct_eq`

thespooler:
I will open a new PR to split argonautica and the rest.

PR #85, opened 2020-04-13
Remove ring from pkce and generator

thespooler:
This improves PR #84 by not including argonautica and only removing ring from pkce and generator.
Also includes changes discussed in #84.

### DIFF
--- a/oxide-auth/Cargo.toml
+++ b/oxide-auth/Cargo.toml
@@ -16,9 +16,13 @@ autoexamples = false
 [dependencies]
 base64 = "0.11"
 chrono = "0.4.2"
+hmac = "0.7.1"
 serde = "1.0"
 serde_derive = "1.0"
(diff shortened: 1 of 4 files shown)


thespooler:
Also, this is a step towards #55

Geobert:
Hi, 

is there a plan to backport the removal of ring in `0.4` branch? 

I'd like to use `oxide-auth` in a production environment but ring version pulled by `oxide-auth` 0.4.5 (through `actix-web` 0.7) is clashing with the one pulled by `warp` 0.2

HeroicKatora:
@Geobert There are no plans. Since the replacement also changed the algorithms used this would also be a breaking change in any case, or require more work than a simple backport.

Geobert:
Thank you for your answer!

I see. I'm looking forward to 0.5 then :D Thanks again!

Issue #89, opened 2020-04-30
[Question and maybe a Bug] What is expected in WebRequest::authheader?


Geobert:
Hi,

I'm trying to authenticate a client using simple::Request.
I'm mapping warp request to simple::Request field and I put this kind of thing in `auth`:
 ```
Request {
    query: {},
    urlbody: {
        "client_id": "toto",
        "grant_type": "authorization_code",
        "redirect_uri": "http://127.0.0.1:19444/callback",
        "code": "e/xuf3B0J+13Urk1KreojQ==",
    },
    auth: Some(
        "Basic dG90bzpwYXNzd29yZA==",
    ),
}
```

but access_token_flow returns:
```
Response {
    status: BadRequest,
    location: None,
    www_authenticate: None,
    body: Some(
        Json(
            "{\"error\":\"invalid_request\"}",
        ),
    ),
}
```
(and I go into `check` without a password if I put `None` in `auth`, so the other fields seem correct)

Regards,

Geobert:
I might have found a bug: https://github.com/HeroicKatora/oxide-auth/blob/abb1f35099a22e46dc1ba6f6c81f85d845d237d4/oxide-auth/src/code_grant/accesstoken.rs#L160

I don't have `client_secret` in the body, so it's `None`.

There's a call to `unauthenticated` but `credentials` was init line 151 so we override the value here, ending in Credentials::Duplicate which turn to an error when calling `into_client`.

Am I right? Or am I missing something?

Geobert:
My request is bad: client_id shouldn't exists in the body when using Basic authorization

PR #90, opened 2020-05-01
fix: Credentials being overriden

Geobert:
This fixes

 - [x] Corresponds to issue (#89)

I'm not sure on how to stub to write a test. But the change seems trivial.

### DIFF
--- a/Contributors.txt
+++ b/Contributors.txt
@@ -4,3 +4,4 @@
 HeroicKatora <andreas.molzer@gmx.de> (initial author)
 ParisLiakos (Paris Liakos, gotham frontend)
 asonix <asonix@asonix.dog> (actix 1.0 update)
+Geobert Quach <geobert.quach@isode.com> (as part of Isode Ltd)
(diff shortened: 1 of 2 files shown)


HeroicKatora:
The change is anything but trivial, this modifies authentication and _must_ have tests for the bug fix. One thing that previously was not allowed, and now is but shouldn't be, is to have both an authentication in the body and an authorization header. This is bad, I remember actively testing this and currently worried why the PR doesn't fail [this test](https://github.com/HeroicKatora/oxide-auth/blob/next/oxide-auth/src/endpoint/tests/authorization.rs#L486).

Geobert:
> 
> 
> The change is anything but trivial, this modifies authentication and _must_ have tests for the bug fix. One thing that previously was not allowed, and now is but shouldn't be, is to have both an authentication in the body and an authorization header. This is bad, I remember actively testing this and currently worried why the PR doesn't fail [this test](https://github.com/HeroicKatora/oxide-auth/blob/next/oxide-auth/src/endpoint/tests/authorization.rs#L486), and why it fails some other unrelated test now whereas the branch came back clean...

Yes, Just seen that, looking into it :)

HeroicKatora:
>EDIT: your link goes nowhere, there's no line 486

Oops, wrong file :smile:

Geobert:
So here are the steps I'm taking:
- look why private_duplicate_authentication doesn't fail and make it fail
- write a test for my bug
- fix the bug

Are you OK with that?

HeroicKatora:
That seems fine :+1:

Geobert:
actually the test doesn't fail and it's right to not to: the Duplicate is properly caught, I understand better what you mean.

(English is not my native language and it's Friday afternoon here ^^')

Geobert:
Oh, maybe there's no bug: 
How should we react if there's client_id in the body (but no client_secret)?

PR #96, opened 2020-07-10
Async access token test

Geobert:
Sorry for the big PR, I tried to split into 2 commits but the 1st one need all this to continue to build.

It's basically implementation of everything needed to run tests for access token.

Then I found some issue so 2nd commit to make the tests work!


oxide-auth-async/src/tests/access_token.rs
```diff
@@ -0,0 +1,669 @@
...
+        AccessTokenEndpoint {
+            registrar,
+            authorizer,
+            issuer,
+            response: Vacant,
+        }
+    }
+}
+
+impl<'a> Endpoint<CraftedRequest> for AccessTokenEndpoint<'a> {
```
HeroicKatora, line 55: I do have one question not directly related to the specific PR here. Did having the define this `impl` feel more cumbersome than using the `simple::Generic` endpoint in the synchronous version? The reason I'm asking is that, in the past several different pieces of feedback hinted that it was hard to discover that you do not _need_ to use `Generic` and that defining a custom trait impl can be much more ergonomic in complex cases. Therefore I was considering moving away from providing `simple::Generic` in the first place, or solely moving it into the tests. I'm just not sure if this removes an interface with usability advantages compared to the 'manual' impl.

oxide-auth-async/src/endpoint/mod.rs
```diff
@@ -49,12 +54,22 @@ pub trait Endpoint<Request: WebRequest> {
...
     ///
     /// Returning `None` is the default implementation and acts as simply providing any extensions.
     fn extension(&mut self) -> Option<&mut dyn Extension> {
         None
     }
-    */
+}
+
+pub trait Extension {
+    /// The handler for authorization code extensions.
```
HeroicKatora, line 66: Please add a `FIXME` to things that are commented out and are somewhat central to the trait or would be a breaking change to introduce.

HeroicKatora:
Also, please add `oxide-auth-async` to [the test CI cases](https://github.com/HeroicKatora/oxide-auth/blob/next/.cirrus.yml#L13) now that it has the first tests.

Geobert:
Thank you for your review :D I'll apply your comment on Monday ;)

oxide-auth-async/src/frontends/mod.rs
```diff
@@ -0,0 +1 @@
+pub mod simple;
```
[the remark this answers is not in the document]
Geobert: Exact :) Should I remove the whole module?

oxide-auth-async/src/tests/access_token.rs
Geobert: Indeed, when using the lib for the first time, I was quite confuse on if I should use `Generic` or not. At the moment, I'm using it in my app, but now that I've done this Endpoint myself, it seems easy enough that I'm considering to move away from Generic :) 

Dealing with generic type is quite hard to track, at least, it is for my brain :D

oxide-auth-async/src/frontends/mod.rs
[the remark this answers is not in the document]
HeroicKatora: Yes, please.

Geobert:
I think I've processed all the comments :)

Geobert:
I also have this branch https://github.com/Geobert/oxide-auth/tree/async-resource

but if I make a PR from it, it will contains the commits from this PR as well as it's not merged yet :) So I won't create the PR yet, but let you know that the branch exists if you want to have a look already :)

PR #97, opened 2020-07-14
Async resource

Geobert:
All needed code for resource flow + tests


HeroicKatora:
@Geobert Would you mind investigating and fixing why `actix` is failing on CI next? I'm somewhat puzzled still and it doesn't appear to be spurious.

Geobert:
> 
> 
> @Geobert Would you mind investigating and fixing why `actix` is failing on CI next? I'm somewhat puzzled still and it doesn't appear to be spurious.

This error you mean?

```
error[E0432]: unresolved import `crossbeam_utils::Backoff`
```

I'll try to have a look, but I don't use actix (I'm using warp) so I don't know if I'll have the knowledge to debug that. I'll try :)

PR #98, opened 2020-07-14
Async refresh

Geobert:
Refresh flow and tests

Also fixed the actix compilation: https://github.com/crossbeam-rs/crossbeam/issues/312


Geobert:
Oh and this PR has `OwnerSolicitor` so you can check the box ;)

oxide-auth-actix/Cargo.toml
```diff
@@ -19,6 +19,7 @@ futures = "0.1"
 oxide-auth = { version = "0.5.0-preview.0", path = "../oxide-auth" }
 serde_urlencoded = "0.5.1"
 url = "1.7"
+crossbeam-utils = "0.6.4"
```
HeroicKatora, line 22: Was this _just_ to get the version updated? I don't see it used anywhere.

Geobert: It's used somewhere in the dep tree. By putting this version on our dep list, we force all 0.6.x version of crossbeam-utils to use this 0.6.4 version.

But I think there was a mistake made by adding a type without bumping the version to 0.7

HeroicKatora: Okay, then would you try either updating or deleting the `Cargo.lock` file? There should be any reason to depend on a specific version directly other than to override a dependency but `cargo` should resolve to the latest version by-default, except if the lock file tells it to do something different.

Geobert: I chosed to delete the lock file as for a library it makes less sense to have one :)

Geobert: Oops, CI `cat` the file… do we need that? If yes I'll update the lock file :)

HeroicKatora: No, it doesn't need to. Respectively, it can also regenerate the file before `cat`ing it

HeroicKatora: > cargo update && cat Cargo.lock

https://github.com/HeroicKatora/static-alloc/blob/master/.cirrus.yml#L6

PR #105, opened 2020-08-18
Async sendable

Geobert:
This contains all I needed to change in order to use it in my project + warning fixes (which included Option of Box instead of Box of Option)

### DIFF
--- a/oxide-auth-async/src/endpoint/mod.rs
+++ b/oxide-auth-async/src/endpoint/mod.rs
@@ -11,7 +11,11 @@ pub mod access_token;
 pub mod refresh;
 pub mod resource;
 
-pub trait Endpoint<Request: WebRequest> {
+pub trait Endpoint<Request>
+where
+    Request: WebRequest + Clone + Send,
+    <Request as WebRequest>::Error: Clone + Send,
+{
     /// The error typed used as the error representation of each flow.
     type Error;
 
--- a/oxide-auth-async/src/endpoint/mod.rs
+++ b/oxide-auth-async/src/endpoint/mod.rs
@@ -62,19 +62,19 @@ pub trait Endpoint<Request: WebRequest> {
     /// Get the central extension instance this endpoint.
     ///
     /// Returning `None` is the default implementation and acts as simply providing any extensions.
-    fn extension(&mut self) -> Option<&mut dyn Extension> {
+    fn extension(&mut self) -> Option<&mut dyn Extension<Request>> {
         None
     }
 }
 
-pub trait Extension {
+pub trait Extension<R: WebRequest> {
     /// The handler for authorization code extensions.
-    fn authorization(&mut self) -> Option<&mut dyn AuthorizationExtension> {
+    fn authorization(&mut self) -> Option<&mut dyn AuthorizationExtension<R>> {
         None
     }
 
     /// The handler for access token extensions.
-    fn access_token(&mut self) -> Option<&mut dyn AccessTokenExtension> {
+    fn access_token(&mut self) -> Option<&mut dyn AccessTokenExtension<R>> {
         None
     }
 }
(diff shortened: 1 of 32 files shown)


Geobert:
Hi @HeroicKatora, 

Do you have an estimate on when you'll have the time to review/merge this? (For my weekly report to my boss :) )

Cheers,

HeroicKatora:
This week for sure, probably tomorrow or Thursday.

Geobert:
Thank you very very much :D

oxide-auth-async/src/endpoint/mod.rs
```diff
@@ -93,14 +95,14 @@ pub trait OwnerSolicitor<Request: WebRequest> {
...
+// impl<T, Request: WebRequest> OwnerSolicitor<Request> for T
+// where
+//     T: oxide_auth::endpoint::OwnerSolicitor<Request> + ?Sized,
+// {
+//     async fn check_consent(
+//         &mut self, req: &mut Request, pre_grant: &PreGrant,
+//     ) -> OwnerConsent<Request::Response> {
+//         oxide_auth::endpoint::OwnerSolicitor::check_consent(self, req, pre_grant)
+//     }
+// }
```
HeroicKatora, line 108: Just noticed this, can this be re-added with the appropriate bounds on `T`?

HeroicKatora (commented):
I'm good on the code. The only thing I worry about is that the requirement of flows to be `Send` is  accidentally broken by some unintended code change, such as reordering yield points? This, I think, requires some additional testing of the future's properties. I've constructed have a small test to ensure this property—will make a PR back to your branch—however it fails on `resource::protect`

<details>
<summary>Full error</summary>
<pre>
error: future cannot be sent between threads safely
   --> oxide-auth-async/src/tests/type_properties.rs:13:5
    |
5   |     fn require_send<T: Send, A, B, F: FnOnce(A, B) -> T>(_: F) {}
    |                        ---- required by this bound in `tests::type_properties::require_futures_have_send_bounds::require_send`
...
13  |     require_send(code_grant::resource::protect);
    |     ^^^^^^^^^^^^ future returned by `protect` is not `Send`
    |
    = help: the trait `std::marker::Send` is not implemented for `dyn primitives::Issuer`
note: future is not `Send` as this value is used across an await
   --> oxide-auth-async/src/code_grant.rs:111:33
    |
111 |                        let grant = handler
    |   _________________________________^
    |  |_________________________________|
    | ||
112 | ||                         .issuer()
    | ||_________________________________- has type `&mut dyn primitives::Issuer` which is not `Send`
113 | |                          .recover_token(&token)
114 | |                          .await
    | |_______________________________^ await occurs here, with `handler
                        .issuer()` maybe used later
115 |                            .map_err(|_| Error::PrimitiveError)?;
    |                                                                - `handler
                        .issuer()` is later dropped here
help: consider moving this into a `let` binding to create a shorter lived borrow
   --> oxide-auth-async/src/code_grant.rs:111:33
    |
111 |                       let grant = handler
    |  _________________________________^
112 | |                         .issuer()
113 | |                         .recover_token(&token)
    | |______________________________________________^
</pre>
</details>

The fix should be easy: 

`resource::Endpoint::issuer` should return `&mut (dyn crate::primitives::Issuer + Send)` like the other traits that return mutable dyn primitives. Can test if that change is compatible with your code base, and if so push the change?

Geobert:
> Can you test if that change is compatible with your code base, and if so push the change?

Thank you for the review! I'll do that asap (on another task at the moment) :)

Geobert:
It's done :D  Still working with my app.

As you noticed, I don't use `resource`, hence the issue… your test was a good call!

HeroicKatora (approved):
Thank you, so much :1st_place_medal:

Geobert:
> 
> 
> Thank you, so much 🥇

You're welcome! And thank you for the review and precious advice!

Issue #116, opened 2020-10-08
RegisteredUrl kind: IgnorePortOnLocalhost


HeroicKatora:
# Feature

This new variant for registered URLs should do _exact_ string matching except where the redirect URL has the host `localhost`, where it should ignore the port. This matches [the latest IETF recommendations](https://www.ietf.org/id/draft-ietf-oauth-security-topics-16.html#name-protecting-redirect-based-f).

## Alternatives

Register a range of alternative redirect url, each only differing in a port.

## Context

When a local app receives a redirect-based callback it must open a web server. Since it does not run in priviledge context it can not predict the available port numbers and typically either let's the system assign any or chooses one from a large number of potentially available ports (such as a range of 10-100 ports).

## Tracking pull request

 - [ ] A pull request does not yet exist

PR #122, opened 2021-04-11
Add IgnoreLocalPortUrl variant for registered URLs

gabriel-araujjo:

 - [x] Corresponds to issue #116 

### DIFF
--- a/oxide-auth/src/primitives/registrar.rs
+++ b/oxide-auth/src/primitives/registrar.rs
@@ -63,6 +63,12 @@ pub enum RegisteredUrl {
     Exact(ExactUrl),
     /// An URL that needs to match the redirect URL semantically.
     Semantic(Url),
+    /// Same as [`Exact`] variant, except where the redirect URL has the host
+    /// `localhost`, where it ignores the port. This matches the [IETF recomendations].
+    ///
+    /// [`Exact`]: Self::Exact
+    /// [IETF recomendations]: https://tools.ietf.org/html/draft-ietf-oauth-security-topics-16#section-2.1
+    IgnorePortOnLocalhost(IgnoreLocalPortUrl),
 }
 
 /// A redirect URL that must be matched exactly by the client.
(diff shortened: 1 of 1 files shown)


HeroicKatora:
I'll review this soon, it looks tidy and useful. Thanks already for the PR :)

HeroicKatora:
(Feel free to ping me if I forget)

HeroicKatora (changes requested):
*from the off*: he forgot

Anyways, my sentiment didn't change. I'm fairly content with the changes except I want to get rid of all the `unsafe`. It shouldn't matter that much for performance and unless you've fuzzed that intensively I don't find it quite convincing enough as the correctness partially depends on `url`'s implementation.

LGTM with the indexing operations changed to indexing with panics.

oxide-auth/src/primitives/registrar.rs
```diff
@@ -327,6 +396,129 @@ impl core::str::FromStr for ExactUrl {
...
+            + match parsed.host_str() {
+                Some(h) => h.len() + 2,
+                None => 0,
+            };
+
+        let port_len = match parsed.port() {
+            None => 0,
+            // SAFITY: Since `url` is a valid URL, it's always an ASCII sequence,
+            // so it's safe to get an uncheck range.
+            Some(_) => match unsafe { url.get_unchecked(port_pos + 1..) }
```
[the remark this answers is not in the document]
gabriel-araujjo: I changed the checking of `IgnoreLocalPortUrl`. Can you please review that last change, please.

oxide-auth/src/primitives/registrar.rs
```diff
@@ -327,6 +390,77 @@ impl core::str::FromStr for ExactUrl {
...
+        let mut parsed: Url = url.parse()?;
+        match parsed.host_str() {
+            Some("localhost") => {
+                let _ = parsed.set_port(None);
+                Ok(IgnoreLocalPortUrl(IgnoreLocalPortUrlInternal::Local(parsed)))
+            }
+            _ => Ok(IgnoreLocalPortUrl(IgnoreLocalPortUrlInternal::Exact(
+                url.into_owned(),
+            ))),
+        }
```
HeroicKatora, line 406: Can you use `From<Url>` here for consistency?

HeroicKatora (commented):
I like this approach a lot better. Just one style nit :)

oxide-auth/src/primitives/registrar.rs
gabriel-araujjo: If I replace that `match` with `Ok(IgnoreLocalPortUrl::from(parsed))`, so `IgnoreLocalPortUrl::new("http://example.com") == IgnoreLocalPortUrl(IgnoreLocalPortUrlInternal::Exact("http://127.0.0.1:8080/".to_string()))` (the normalized form of `"http://example.com"`).

That change can lead to strange behavior, like `RegisteredUrl::IgnorePortOnLocalhost(IgnoreLocalPortUrl::new("http://example.com"))` being different of `RegisteredUrl::Exact("http://example.com".to_string())`, but equal to `RegisteredUrl::Exact("http://example.com/".to_string())`.

HeroicKatora: I see, there is a subtle difference with where `url.into_owned()` would become `parsed.into()` if changed. It irks me a bit that the main logic—the part that is also critical for fitness for purpose—of matching the host and unsetting the port is not obviously shared between those two. What a shame that `Url::parse().to_string()` does not roundtrip.

Okay, consider this solved then but feel free to add more comments explaining that subtlety in case a later programmer intends on changing and consolidating the two portions.

HeroicKatora (approved):

PR #127, opened 2021-09-18
Implement token rotation for `TokenMap`

danielalvsaaker:
This improves the security of the token refresh flow for the `TokenMap` primitive by returning a new refresh token every time, invalidating the previous one.

As mentioned here: https://datatracker.ietf.org/doc/html/rfc6819#section-5.2.2.3

### DIFF
--- a/oxide-auth/src/primitives/issuer.rs
+++ b/oxide-auth/src/primitives/issuer.rs
@@ -635,6 +635,23 @@ pub mod tests {
     }
 
     #[test]
+    fn random_refresh_rotation() {
+        let mut token_map = TokenMap::new(RandomGenerator::new(16));
+        let issued = token_map.issue(grant_template());
+
+        let token = issued.expect("Issuing with refresh token failed");
+        let refresh = token.refresh.expect("No refresh token returned");
+
+        let refreshed_token = token_map
+            .refresh(&refresh, grant_template())
+            .expect("Failed to refresh access token");
+
+        let new_refresh = refreshed_token.refresh.expect("No new refresh token returned");
+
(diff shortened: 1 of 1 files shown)


oxide-auth/src/primitives/issuer.rs
```diff
@@ -287,9 +287,13 @@ impl<G: TagGrant> Issuer for TokenMap<G> {
         self.set_duration(&mut grant);
         let until = grant.until;
 
-        let next_usage = self.usage.wrapping_add(1);
+        let next_usage = self.usage.wrapping_add(2);
+
         let new_access = self.generator.tag(self.usage, &grant)?;
-        let new_key: Arc<str> = Arc::from(new_access.clone());
+        let new_refresh = self.generator.tag(self.usage.wrapping_add(1), &grant)?;
```
HeroicKatora, line 293: From a semantic standpoint this is fine. From a security standpoint it could be cleaner to more clearly show that you're using two different `usage` tags and that the `next_usage` will also differ. I was thinking either a comment or assigning the two cases to variables first.

HeroicKatora (commented):
I usually wouldn't ask for changes on totally good code but for the security critical portion it seems right to do one work of structurual improving on every change :slightly_smiling_face:

oxide-auth/src/primitives/issuer.rs
danielalvsaaker: @HeroicKatora Made a change now, assigned tags to a variable between token generation. Let me know if this is what you had in mind.

HeroicKatora (approved):
LGTM! code flow is linear enough to be obvious to me now.

Issue #131, opened 2021-12-30
Upgrade rocket to 0.5


Maweypeyyu:
# Project Improvement

Locally swapped out code to work with the rocket 0.5-rc1 and thought why not share it. I understand that this might not be included right now, although I did update the version.

# Tracking pull request

 - [x] A pull request is being created after this issue.

jtmorrisbytes:
I have made a second attempt to upgrade to rocket 0.5. Some issues that arrived are as follows:

1. rocket 0.5 went from Sync -> Async code. As such I had to work around serde_urlencoded since rocket::data::DataStream does not implement std::io::Read anymore and instead implements tokio::io::AsyncRead
2. I had to introduce tokio as a dependency to this crate using just the io-util feature
3.  rocket introduced data limits when opening a raw datastream. I had to modify OAuthRequest::add_body to take an additional optional parameter called 'limits' that passed along the server configured limits. if no limit is specified, then the add_body method will use the hard coded default provided by rocket
4. I had difficulty with lifetimes in the FromRequest implementation. It is difficult to use borrowed data in FromRequest because you somehow have to specify that the lifetime of the OAuthRequest structure lives as long as the request structure but you are returning a new instance of OAuthRequest with borrowed data. I tried a couple of patterns and the only one that seemed to work is where `impl<'r> FromRequest<'r> for OAuthRequest<'static>`. the examples seem to compile on my machine except for the support module due to the #[path} statement. I still need to work on adding tests and updating docs.

This is my first time contributing to another project. please let me know if there is anything that needs to be done here and I will submit another pull request

icode:
Is it still not usable?

PR #132, opened 2021-12-30
upgrades rocket to 0.5

Maweypeyyu:

 - [x] Corresponds to issue (#131)

Let me know if I can improve anything.

### DIFF
--- a/oxide-auth-rocket/src/lib.rs
+++ b/oxide-auth-rocket/src/lib.rs
@@ -6,9 +6,9 @@ mod failure;
 use std::io::Cursor;
 use std::marker::PhantomData;
 
-use rocket::{Data, Request, Response};
-use rocket::http::{ContentType, Status};
-use rocket::http::hyper::header;
+use rocket::data::ToByteUnit;
+use rocket::{Data, Request, Response, uri};
+use rocket::http::{ContentType, Status, Header};
 use rocket::request::FromRequest;
 use rocket::response::{self, Responder};
 use rocket::outcome::Outcome;
@@ -31,7 +31,7 @@ pub struct OAuthRequest<'r> {
 /// Response type for Rocket OAuth requests
 ///
 /// A simple wrapper type around a simple `rocket::Response<'r>` that implements `WebResponse`.
-#[derive(Debug)]
+#[derive(Debug, Default)]
 pub struct OAuthResponse<'r>(Response<'r>);
 
 /// Request error at the http layer.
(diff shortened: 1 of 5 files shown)


oxide-auth-rocket/src/lib.rs
```diff
@@ -58,9 +58,11 @@ impl<'r> OAuthRequest<'r> {
     /// Create the request data from request headers.
     ///
     /// Some oauth methods need additionally the body data which you can attach later.
-    pub fn new<'a>(request: &'a Request<'r>) -> Self {
-        let query = request.uri().query().unwrap_or("");
-        let query = match serde_urlencoded::from_str(query) {
+    pub fn new(request: &Request<'_>) -> Self {
+        let default_query_uri = uri!("?b");
+        let default_query = default_query_uri.query().unwrap();
+        let query = request.uri().query().unwrap_or(default_query);
```
HeroicKatora, line 64: This will need some justification/explanation as in additional comments. Why this specific query string instead of the empty one previously? Is there some type change that requires this—in a first look it seems that maybe a simple clone `&query.to_string()` would have sufficed? Justifiy the unwraps as well, please.

oxide-auth-rocket/src/lib.rs
```diff
@@ -94,12 +96,13 @@ impl<'r> OAuthRequest<'r> {
...
     /// type for both these use cases. When you forget to provide the body to a request, the oauth
     /// system will return an error the moment the request is used.
-    pub fn add_body(&mut self, data: Data) {
+    pub async fn add_body(&mut self, data: Data<'_>) {
         // Nothing to do if we already have a body, or already generated an error. This includes
         // the case where the content type does not indicate a form, as the error is silent until a
         // body is explicitely requested.
         if let Ok(None) = self.body {
-            match serde_urlencoded::from_reader(data.open()) {
+            let data = data.open(2.mebibytes()).into_string().await;
```
HeroicKatora, line 104: Arbitrary limit should be a constant somewhere, or, even better, be configurable on the object for the caller. Probably both where a default is set with a constant and it can be overwritten.

HeroicKatora (commented):
Nice, looks like the next rocket version, when it is finally released…, will provide some interesting asynchronous wins. I wonder if it could be moved to `oxide-auth-async` even. Don't worry about that latter part in this PR though, that would be much harder to review.

oxide-auth-rocket/src/lib.rs
Maweypeyyu: Will try more tomorrow, there have been changes to uri/query that broke the previous code.

HeroicKatora:
Since  4 months ago, `rocket = 0.5` has been released. I'd look into a rebase at some point, thanks for the sketches of the expected changes with the pre-releases.

Maweypeyyu:
Work took over and I hadn't had a chance to look into this since then. Sorry about that.

HeroicKatora:
Please don't worry about it. Same for me, and rocket dragged that out quite a bit as well. It's FOSS because we appreciate what you do, however far it goes :smile:

PR #135, opened 2022-02-28
Release oxide auth actix 0.2

HeroicKatora:


thespooler:
If the release step ended in success, does it mean that the package should've been published to crates.io?

HeroicKatora:
No, every publish action is done manually. It's just an extra validation step that's only performed on release branches.

PR #136, opened 2022-04-08
Derive Serialize/Deserialize for `EncodedClient`

danielalvsaaker:
This adds support for custom implementations of `Registrar` to store encoded clients in a serialized format of choice.

-----------------------------------

### DIFF
--- a/oxide-auth/src/primitives/registrar.rs
+++ b/oxide-auth/src/primitives/registrar.rs
@@ -100,7 +100,7 @@ impl<'de> Deserialize<'de> for ExactUrl {
     where
         D: serde::Deserializer<'de>,
     {
-        let string = String::deserialize(deserializer)?;
+        let string: &str = Deserialize::deserialize(deserializer)?;
         core::str::FromStr::from_str(&string).map_err(serde::de::Error::custom)
     }
 }
@@ -145,8 +145,8 @@ impl<'de> Deserialize<'de> for IgnoreLocalPortUrl {
     where
         D: serde::Deserializer<'de>,
     {
-        let string = String::deserialize(deserializer)?;
-        Self::new(&string).map_err(serde::de::Error::custom)
+        let string: &str = Deserialize::deserialize(deserializer)?;
+        Self::new(string).map_err(serde::de::Error::custom)
     }
 }
 
--- a/oxide-auth/src/primitives/registrar.rs
+++ b/oxide-auth/src/primitives/registrar.rs
@@ -16,6 +16,7 @@ use std::sync::{Arc, MutexGuard, RwLockWriteGuard};
 use argon2::{self, Config};
 use once_cell::sync::Lazy;
 use rand::{RngCore, thread_rng};
+use serde::{Deserialize, Serialize};
 use url::{Url, ParseError as ParseUrlError};
 
 /// Registrars provie a way to interact with clients.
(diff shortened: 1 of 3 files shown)


oxide-auth/src/primitives/registrar.rs
```diff
@@ -91,7 +92,7 @@ pub enum RegisteredUrl {
 /// possible if clients are allowed to provide any semantically matching URL as there are
 /// infinitely many with different hashes. (Note: a hashed form of URL storage is not currently
 /// supported).
-#[derive(Clone, Debug, PartialEq, Eq)]
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
 pub struct ExactUrl(String);
```
HeroicKatora, line 96: There's an invariant for this field: it _must_ represent a valid URL. It's not memory-safety relevant, however, violating this can cause crashes later on as the URL is parsed with unwrap later. `serde` deserialization should check this property.

oxide-auth/src/primitives/registrar.rs
```diff
@@ -120,10 +121,10 @@ pub struct ExactUrl(String);
 /// considered.
 ///
 /// [`ExactUrl`]: ExactUrl
-#[derive(Clone, Debug, PartialEq, Eq)]
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
 pub struct IgnoreLocalPortUrl(IgnoreLocalPortUrlInternal);
```
HeroicKatora, line 125: Similar issue here as for `ExactUrl` but here it's actually _safety relevant_ that the `Local`-variant must not be constructed from a non-`localhost` domain. Deserialization _must_ critically go through `IgnoreLocalPortUrl::new` for this struct.

HeroicKatora (changes requested):
I don't mind serde-deriving for all fully public structs (feel free to add those for all the other primitives as well, there's a fork that added a different set of derives).

However, for structs with non-public members it's much less clear. Both Scope and URL-variants represent (url-)strings and possibly should be serialized as such. More critically, the structs have correctness, safety, and security invariants and `derive(Deserialize)` allows anyone to circumvent them—quite possibly with attacker controlled data. Actually, very likely with attacker controlled data if you follow a Defense in Depth train of thought. Please make serialization go through the public constructors that perform the necessary safety checks.

danielalvsaaker:
You're completely right, and I appreciate the discussion/feedback. I pushed a commit based on your comments, let me know if this is what you had in mind.

oxide-auth/src/primitives/registrar.rs
```diff
@@ -120,15 +131,25 @@ pub struct ExactUrl(String);
 /// considered.
 ///
 /// [`ExactUrl`]: ExactUrl
-#[derive(Clone, Debug, PartialEq, Eq)]
+#[derive(Clone, Debug, PartialEq, Eq, Serialize)]
```
HeroicKatora, line 134: In the current implementation, `Serialize` is not the inverse of `Deserialize`.

HeroicKatora (commented):
Could you add roundtrip tests to ensure that the non-derived `Serialize`/`Deserialize` pairs?

danielalvsaaker:
> Could you add roundtrip tests to ensure that the non-derived `Serialize`/`Deserialize` pairs?

Added tests for both roundtrip serialization and deserialization from invalid input.

HeroicKatora (approved):

HeroicKatora:
Thank you for these, release coming soon :+1:

PR #138, opened 2022-04-18
Add Poem Integration

l1npengtul:
This adds Poem support to oxide-auth.


oxide-auth-poem/src/request.rs
```diff
@@ -0,0 +1,90 @@
...
+            .ok()
+            .map(|f| f.0);
+
+        let mut all_auth = req.headers().get_all("Authorization").into_iter();
+        let optional = all_auth.next();
+
+        let auth = match all_auth.next() {
+            Some(_) => return Err(BadRequest(OxidePoemError::Authorization)),
+            None => optional.and_then(|header| header.to_str().ok().map(str::to_owned)),
+        };
```
HeroicKatora, line 86: Thanks for the gentle reminder that `oxide-auth`'s authorization API requires implementations to catch this, and fail on requests with duplicate headers, as well as enforces `str` improperly. Looks good; I'll convert it to an issue with `oxide-auth` after accepting the PR.

oxide-auth-poem/src/error.rs
```diff
@@ -0,0 +1,11 @@
+use thiserror::Error;
+
+#[derive(Clone, Debug, Error)]
+pub enum OxidePoemError {
+    #[error("There was a problem with the request")]
```
HeroicKatora, line 5: This _could_ be more specific. Something along the lines of 'Invalid request query' and 'Invalid request body' with more precise error information would also work.

HeroicKatora (approved):
Looks good, a few comments that aren't critical to release.

How do you want to handle publishing? In any case a busfactor of two should be beneficial for bindings that I have no personal stake in. You can publish and send an invite on crates.io or do it the other way around.

oxide-auth-poem/src/error.rs
l1npengtul: Yep, will work on that.

l1npengtul:
Just fixed a couple of things, and added a few clippy lints. 

Ill publish and invite you over once merged. 

Thanks for reviewing my PR!

l1npengtul:
I sent over an invite on crates.io

PR #140, opened 2022-05-04
Fix clippy warnings

l1npengtul:
This PR fixes ~40+ clippy warnings and documentation typos in the main crate, along with many others in the subsequent crates for those using `oxide-auth`.

This also bumps oxide-auth and all its related crates by 1 minor version to comply with SemVer(lots of `Into -> From`s, possible breaking API changes) and to update the oxide-auth to 0.6

### DIFF
--- a/oxide-auth-rouille/Cargo.toml
+++ b/oxide-auth-rouille/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "oxide-auth-rouille"
-version = "0.1.0"
+version = "0.1.1"
 authors = ["Andreas Molzer <andreas.molzer@gmx.de>"]
 repository = "https://github.com/HeroicKatora/oxide-auth.git"
 edition = "2018"
@@ -13,7 +13,7 @@ license = "MIT OR Apache-2.0"
 
 [dependencies]
 rouille = "3.0"
-oxide-auth = { version = "0.5.0", path = "../oxide-auth" }
+oxide-auth = { version = "0.6", path = "../oxide-auth" }
 serde_urlencoded = "0.7"
 url = "2"
 
--- a/oxide-auth/src/code_grant/authorization.rs
+++ b/oxide-auth/src/code_grant/authorization.rs
@@ -70,7 +70,7 @@ pub trait Endpoint {
     fn extension(&mut self) -> &mut dyn Extension;
 }
 
-/// The result will indicate wether the authorization succeed or not.
+/// The result will indicate whether the authorization succeed or not.
 pub struct Authorization {
     state: AuthorizationState,
     extensions: Option<Extensions>,
@@ -136,7 +136,7 @@ pub enum Output<'machine> {
     },
     /// Ask for extensions if any
     Extend,
-    /// Ask registrar to negociate
+    /// Ask registrar to negotiate
     Negotiate {
         /// The current bound client
         bound_client: &'machine BoundClient<'static>,
@@ -170,7 +170,7 @@ impl Authorization {
     }
 
     /// Go to next state
-    pub fn advance<'req>(&mut self, input: Input<'req>) -> Output<'_> {
+    pub fn advance(&mut self, input: Input) -> Output {
         self.state = match (self.take(), input) {
             (current, Input::None) => current,
             (
@@ -207,7 +207,7 @@ impl Authorization {
             },
             AuthorizationState::Extending { .. } => Output::Extend,
             AuthorizationState::Negotiating { bound_client } => Output::Negotiate {
-                bound_client: &bound_client,
+                bound_client,
                 scope: self.scope.clone(),
             },
             AuthorizationState::Pending {
@@ -216,7 +216,7 @@ impl Authorization {
                 extensions,
             } => Output::Ok {
                 pre_grant: pre_grant.clone(),
-                state: state.clone(),
+                state: state.as_ref().map(Clone::clone), // ??
                 extensions: extensions.clone(),
             },
         }
@@ -302,7 +302,7 @@ impl Authorization {
 
 /// Retrieve allowed scope and redirect url from the registrar.
 ///
-/// Checks the validity of any given input as the registrar instance communicates the registrated
+/// Checks the validity of any given input as the registrar instance communicates the registered
 /// parameters. The registrar can also set or override the requested (default) scope of the client.
 /// This will result in a tuple of negotiated parameters which can be used further to authorize
 /// the client by the owner or, in case of errors, in an action to be taken.
@@ -425,9 +425,9 @@ pub fn authorization_code(handler: &mut dyn Endpoint, request: &dyn Request) -> 
 }
 
 /// Represents a valid, currently pending authorization request not bound to an owner. The frontend
-/// can signal a reponse using this object.
+/// can signal a response using this object.
 // Don't ever implement `Clone` here. It's to make it very
-// hard for the user toaccidentally respond to a request in two conflicting ways. This has
+// hard for the user to accidentally respond to a request in two conflicting ways. This has
 // potential security impact if it could be both denied and authorized.
 pub struct Pending {
     pre_grant: PreGrant,
@@ -506,7 +506,7 @@ pub enum Error {
 
 /// Encapsulates a redirect to a valid redirect_uri with an error response. The implementation
 /// makes it possible to alter the contained error, for example to provide additional optional
-/// information. The error type should not be altered by the frontend but the specificalities
+/// information. The error type should not be altered by the frontend but the specifics
 /// of this should be enforced by the frontend instead.
 #[derive(Clone)]
 pub struct ErrorUrl {
@@ -565,11 +565,10 @@ impl Error {
     }
 }
 
-impl Into<Url> for ErrorUrl {
-    /// Finalize the error url by saving its parameters in the query part of the redirect_uri
-    fn into(self) -> Url {
-        let mut url = self.base_uri;
-        url.query_pairs_mut().extend_pairs(self.error.into_iter());
+impl From<ErrorUrl> for Url {
+    fn from(err_url: ErrorUrl) -> Self {
+        let mut url = err_url.base_uri;
+        url.query_pairs_mut().extend_pairs(err_url.error.into_iter());
         url
     }
 }
--- a/oxide-auth/src/code_grant/extensions/pkce.rs
+++ b/oxide-auth/src/code_grant/extensions/pkce.rs
@@ -1,10 +1,45 @@
 use std::borrow::Cow;
+use std::error::Error;
+use std::fmt::{Display, Formatter};
 
-use crate::primitives::grant::{GrantExtension, Value};
+use crate::{
+    primitives::grant::{GrantExtension, Value},
+    util::avoid_alloc_cow_str_to_string,
+};
 
 use base64;
 use sha2::{Digest, Sha256};
 use subtle::ConstantTimeEq;
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum PkceError {
+    NoChallenge,
+    NoVerifier,
+    FailedVerification(&'static str),
+    InvalidMethod(String),
+    UnsupportedMethod,
+}
+
+impl Display for PkceError {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        write!(
+            f,
+            "{}",
+            match self {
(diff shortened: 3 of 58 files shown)


oxide-auth/src/code_grant/authorization.rs
```diff
@@ -216,7 +216,7 @@ impl Authorization {
                 extensions,
             } => Output::Ok {
                 pre_grant: pre_grant.clone(),
-                state: state.clone(),
+                state: state.as_ref().map(Clone::clone), // ??
```
HeroicKatora, line 219: Yes, indeed, `// ??`

oxide-auth/src/code_grant/extensions/pkce.rs
```diff
@@ -128,19 +169,19 @@ fn b64encode(data: &[u8]) -> String {
...
 
 impl Method {
-    fn from_parameter(method: Cow<str>, challenge: Cow<str>) -> Result<Self, ()> {
+    fn from_parameter(method: impl AsRef<str>, challenge: impl AsRef<str>) -> Result<Self, PkceError> {
         match method.as_ref() {
-            "plain" => Ok(Method::Plain(challenge.into_owned())),
-            "S256" => Ok(Method::Sha256(challenge.into_owned())),
-            _ => Err(()),
+            "plain" => Ok(Method::Plain(challenge.as_ref().to_string())),
+            "S256" => Ok(Method::Sha256(challenge.as_ref().to_string())),
```
HeroicKatora, line 175: Now this always performs an allocation and is instantiated multiple times.

oxide-auth/src/code_grant/mod.rs
```diff
@@ -27,7 +27,7 @@
 //! [`endpoint`]: ../endpoint/index.html
 //! [`Endpoint`]: ../endpoint/trait.Endpoint.html
 
-pub mod accesstoken;
+pub mod access_token;
```
HeroicKatora, line 30: Note to self: Breaking

oxide-auth/src/endpoint/access_token.rs
```diff
@@ -225,14 +242,15 @@ impl<'a, R: WebRequest + 'a> WrappedRequest<'a, R> {
...
+            body: Cow::Owned(NormalizedParameter::default()),
             authorization: None,
             error: Some(err),
             allow_credentials_in_body: false,
         }
     }
 
-    fn parse_header(header: Cow<str>) -> Result<Authorization, Invalid> {
+    fn parse_header(header: impl AsRef<str>) -> Result<Authorization, Invalid> {
+        let header = header.as_ref();
```
HeroicKatora, line 253: Can't say I get why this change would be necessary.

oxide-auth/src/endpoint/mod.rs
```diff
@@ -333,7 +347,7 @@ pub trait WebResponse {
     /// A pure text response with no special media type set.
     fn body_text(&mut self, text: &str) -> Result<(), Self::Error>;
 
-    /// Json repsonse data, with media type `aplication/json.
+    /// Json repsonse data, with media type `aplication/json`.
```
HeroicKatora, line 350: Just now noticing another type here: `application/json` is missing a p.

oxide-auth/src/primitives/scope.rs
```diff
@@ -72,20 +72,20 @@ impl Scope {
     fn invalid_scope_char(ch: char) -> bool {
         match ch {
             '\x21' => false,
-            ch if ch >= '\x23' && ch <= '\x5b' => false,
-            ch if ch >= '\x5d' && ch <= '\x7e' => false,
-            ' ' => false, // Space seperator is a valid char
+            ch if ('\x23'..='\x5b').contains(&ch) => false,
+            ch if ('\x5d'..='\x7e').contains(&ch) => false,
```
HeroicKatora, line 76: Note to self: MSRV bump

oxide-auth/src/util.rs
```diff
@@ -0,0 +1,9 @@
+use std::borrow::Cow;
+
+// avoids allocation on `Cow::Owned`.
+pub fn avoid_alloc_cow_str_to_string(cow: Cow<str>) -> String {
+    match cow {
+        Cow::Borrowed(b) => b.to_string(),
+        Cow::Owned(o) => o,
+    }
+}
```
HeroicKatora, line 9: This is just `Cow::into_owned`, is it not?

oxide-auth-rouille/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "oxide-auth-rouille"
-version = "0.1.0"
+version = "0.1.1"
```
HeroicKatora, line 3: If the `oxide-auth` dependency is publicly visible, including as trait-`impl`s, then this must be `0.2`. (Obviously this is the case as implementing the traits is half the reason to have this crate).

oxide-auth/src/code_grant/extensions/pkce.rs
```diff
@@ -1,11 +1,46 @@
...
 
 use base64;
 use sha2::{Digest, Sha256};
 use subtle::ConstantTimeEq;
 
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum PkceError {
+    NoChallenge,
+    NoVerifier,
+    FailedVerification(&'static str),
```
HeroicKatora, line 18: Rule of thumb, don't elaborate on verification failures in user facing context. They failed, everything beyond is a potentialy leak of information. Even if these are only ever filled with one string per method, that's not clear to a  reviewer at this point.

Should be `FailedVerification`, potentially you can add the method in some way but I don't really think it's necessary. Also, can you list this variant last?

oxide-auth/src/code_grant/extensions/pkce.rs
```diff
@@ -151,24 +192,32 @@ impl Method {
...
+                Ok(Method::Plain(string))
+            }
+            Some('S') => {
+                let mut string = avoid_alloc_cow_str_to_string(encoded);
+                string.pop();
+                Ok(Method::Sha256(string))
+            }
+            _ => Err(PkceError::InvalidMethod(
+                "No/invalid encoding character indicator at end of string, expected `p` or `S`"
+                    .to_string(),
```
HeroicKatora, line 209: Explanatory string should be part of the `Display` impl, not a string formatting.

oxide-auth/src/code_grant/extensions/pkce.rs
```diff
@@ -1,11 +1,46 @@
...
+    primitives::grant::{GrantExtension, Value},
+    util::avoid_alloc_cow_str_to_string,
+};
 
 use base64;
 use sha2::{Digest, Sha256};
 use subtle::ConstantTimeEq;
 
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum PkceError {
```
HeroicKatora, line 15: Even though I'm not yet certain about a contentful error type, surely not as an enum. It should be an opaque struct and this enum its `inner: ` attribute.

oxide-auth/src/code_grant/extensions/pkce.rs
```diff
@@ -1,11 +1,46 @@
...
+            "{}",
+            match self {
+                // perhaps use thiserror so we can avoid these allocs?
+                PkceError::NoChallenge =>
+                    "PKCE: It was indicated that an error was required but none was given".to_string(),
+                PkceError::NoVerifier =>
+                    "PKCE: The challenge was agreed upon but there is no verifier present".to_string(),
+                PkceError::FailedVerification(reason) => format!("PKCE: Failed to verify PKCE method: {reason}"),
+                PkceError::InvalidMethod(method) => format!("PKCE: Method must be either `S256`(sha256) or `plain`, but was given: {method}"),
+                PkceError::UnsupportedMethod => "PKCE: Method is unsupported due to indicator `allow_plain: false` and `method: plain`".to_string(),
```
HeroicKatora, line 36: Then the variant should be named `UnsupportedMethodPlain`. Also, don't refer to server internals in user-facing errors. This is fine as a Debug message but less so as Display.

oxide-auth/src/endpoint/mod.rs
```diff
@@ -203,17 +206,20 @@ impl<'flow> Solicitation<'flow> {
...
+    #[must_use]
     pub fn state(&self) -> Option<&str> {
-        match self.state {
-            None => None,
-            Some(ref state) => Some(&state),
-        }
+        // match self.state {
+        //     None => None,
+        //     Some(ref state) => Some(state),
+        // }
```
HeroicKatora, line 214: Code to be removed.

HeroicKatora (commented):
I can't guarantee that my review was complete. I might revisit it and add more comments, later. This would be easier if you separated

- style changes
- SemVer relevant changes as part of naming
- functional changes such as Pkce's error

The first category is always welcome, if you decide to split the PR then we can bring this quickly and in the process reduce the diff for all remaining changes.

oxide-auth/src/endpoint/access_token.rs
l1npengtul: to be honest i probably refactored that part at 3AM and did random things, looking back this is dumb

will change back

oxide-auth/src/code_grant/authorization.rs
```diff
@@ -170,7 +172,7 @@ impl Authorization {
     }
 
     /// Go to next state
-    pub fn advance<'req>(&mut self, input: Input<'req>) -> Output<'_> {
+    pub fn advance(&mut self, input: Input) -> Output {
```
HeroicKatora, line 175: I don't actually know for certain if this is breaking. At face value it looks like this is actually exactly the same. [Reference](https://doc.rust-lang.org/stable/reference/lifetime-elision.html)

- Each elided lifetime in the parameters becomes a distinct lifetime parameter.
- If the receiver has type &Self or &mut Self, then the lifetime of that reference to Self is assigned to all elided output lifetime parameters.

oxide-auth/src/primitives/scope.rs
```diff
@@ -72,20 +72,20 @@ impl Scope {
...
+            ' ' => false, // Space separator is a valid char
             _ => true,
         }
     }
 
     /// Determines if this scope has enough privileges to access some resource requiring the scope
     /// on the right side. This operation is equivalent to comparison via `>=`.
-    pub fn priviledged_to(&self, rhs: &Scope) -> bool {
+    pub fn privileged_to(&self, rhs: &Scope) -> bool {
```
HeroicKatora, line 84: Breaking, but can also be solved with a hidden/deprecated method. Weirdly, I have a faint memory of removing this at some point but it may have been somewhere else.

oxide-auth/src/code_grant/refresh.rs
```diff
@@ -237,7 +237,7 @@ impl Refresh {
     ///
     /// The provided `Input` needs to fulfill the *previous* `Output` request. See their
     /// documentation for more information.
-    pub fn advance<'req>(&mut self, input: Input<'req>) -> Output<'_> {
+    pub fn advance(&mut self, input: Input) -> Output { // Breaking change for 0.6
```
HeroicKatora, line 240: Not actually breaking, I think. This should be the same lifetimes as implied, no?

oxide-auth/src/code_grant/mod.rs
HeroicKatora: We can offer a transitional deprecation by adding an alias mod first:

```rust
#[deprecated = "Switch to the new spelling, access_token"]
pub mod accesstoken {
    pub use super::access_token::*;
}
```

HeroicKatora (commented):
I'm dragging my feet on this a small bit since breaking changes seem a bit unfortunate for pure warnings. I'd really rather the breaking changes were a separate PR; and there'd be one last as cleaned-up as possible release. Consider the main goal being security and that updates will reach downstream _slowly_, the best we can do to speed it up is adding good migration helps but given that the crate was essentially stable for over a year I wouldn't expect fast downstream updates either.

In particular, I aim to have a migration guide covering all breaking changes and it'd be easier to get into the details with separate changes.

Specific actions:
- Factoring out the `PkceError` change to a separate PR
- Providing a deprecated and hidden alias for `mod accesstoken`
- Clarity on the two lifetime changes
- Providing a hidden, deprecated, temporary alias for the wrong spelling `privileged_to`

l1npengtul:
For the unit error warning, I'll just put `#[allow(...)]` blocks in front of all of them, should clear it up nicely.

PR #144, opened 2022-06-28
make "Bearer" handling 'case insensitive'

bbigras:
https://next-auth.js.org/ seems to use a lower case `bearer` which didn't work with oxide-auth.

### DIFF
--- a/oxide-auth/src/code_grant/resource.rs
+++ b/oxide-auth/src/code_grant/resource.rs
@@ -266,7 +266,10 @@ fn validate(request: &'_ dyn Request) -> Result<ResourceState> {
         }
     };
 
-    if !client_token.starts_with(BEARER_START) {
+    if !client_token
+        .to_uppercase()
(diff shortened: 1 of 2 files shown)


HeroicKatora (commented):
Couldn't resist looking up what the RFC says :smile: 

>    HTTP provides a simple challenge-response authentication mechanism
   that MAY be used by a server to challenge a client request and by a
   client to provide authentication information. It uses an extensible,
   case-insensitive token to identify the authentication scheme,
> -- https://datatracker.ietf.org/doc/html/rfc2617#section-1.2

Could you add a test case to `oxide-auth/src/endpoint/tests/resource.rs`? Copying [this one](https://github.com/HeroicKatora/oxide-auth/blob/653853256f863b22811203b77968bec1c742380a/oxide-auth/src/endpoint/tests/resource.rs#L86-L95) with a modified header would be fine :)

PR #146, opened 2022-07-08
Fix 'Cannot drop a runtime' error

olegchir:
This commit fixes immediate client crash in the interactive actix-example.

I used the solution provided in another similar [reqwest bug](https://github.com/seanmonstar/reqwest/issues/1017#issuecomment-1157260218).

It appears that blocking and async do not mix, as reqwest's [documentation](https://github.com/seanmonstar/reqwest/pull/1159) said.

 - [+] I have read the contribution guidelines
 - [-] This change *has no tests*
 - [-] This change *has no documentation*


HeroicKatora:
Thanks! Seems like a no brainer to merge this, although I'm honestly bewildered how this issue could be considered acceptable behavior in the reqwest/tokio combination. Not like blocking is ever *incorrect* (even less so from a pure type standpoint) so I'm unsure why it leads to outright panic.

If you're looking for further contributions then you might be interested in reworking examples to full asynchronous code. Most were created when `async fn` was still young so the style is almost certainly no longer considered idiomatic.

olegchir:
> If you're looking for further contributions then you might be interested in reworking examples to full asynchronous code. 

I'm a total noob in Rust. Coz I'm a backend Java developer who's trying to move to Rust in some performance-critical applications.

I'll try to fix the errors I'll find, but I surely will need your review, coz at the moment I can't tell a difference between idiomatic and non-idiomatic code :)

Issue #125, opened 2021-06-21
Does this library support client credentials grant type?


videni:
Does this library support [OAuth 2.0 Client Credentials Grant Type](https://oauth.net/2/grant-types/client-credentials/) please? I didn't find any doc about this ,

videni:
I check the soure code, it seems not supported yet.

videni:
@HeroicKatora , the `client credentials grant type` might be the easiest one in oauth2,  are you interested?

HeroicKatora:
Without the complexity of the grant type, yes, I would be interested. It should be possible to re-use the existing primitives for this, i.e. a combination of issuer and registrar.

videni:
Are you willing to implement it? otherwise, I will add it later.

HeroicKatora:
I'm afraid I won't have the time to do so myself in the nearby future. However, I'd be willing to provide some mentoring and code review on a PR.

danielalvsaaker:
I'm also interested in this, and wondering how this is best implemented. Should `AccessToken`be generic, swapping the inner state machine depending on the grant type?

HeroicKatora:
The existing backend algorithms shouldn't become generic, I think this would vastly overcomplicate their analysis and testing. I would approach the problem this way:

1. Create a new state machine in `code_grant::accesstoken` for the ClientCredentials kind. Note that input and output of that machine isn't quite the same because we need to handle `scopes` that are otherwise in `authorization` and also the extension module would work differently.
2. Add some explicit new builder (or constructor) to the `AccessTokenFlow` that allows dispatching to that new state machine.

foxfriends:
I took a brief look into what it would take to build the client credentials flow, and ran into one main roadblock that the `Issuer::issue` requires a `Grant` (and the `Registrar` a `PreGrant`) which includes stuff like  `redirect_uri` and `owner_id`. No expert at OAuth, only started playing with it last week, but if I'm not misunderstanding, those seem to be irrelevant for the client credentials flow?

Additionally `Registrar::negotiate` requires the `BoundClient`, but it seems in the client credentials flow, we won't have that either, just a `client_id` which may affect the scopes you are allowed to ask for.

Any thoughts on how that might be handled? A new set of `Issuer` and `Registrar`-like traits for client credentials support that use different types than `Grant`/`PreGrant`, or just making the existing `Issuer`/`Registrar` less strict about what is passed in?

[...]

PR #148, opened 2022-08-19
Client Credentials grant type

foxfriends:
This implements the Client Credentials grant type as discussed in #125. Most of it seemed pretty straightforward, but a few points may be worth discussing...

Following your suggestions, I reused what types I could from `code_grant/accesstoken.rs` and created new versions of the distinct ones in `code_grant/client_credentials.rs`. This was a lot of code, so I put it in a new file instead of just adding it to `accesstoken` as mentioned.

The `AccessTokenFlow` didn't feel like it needed to be complicated with the client credentials, so I just made a separate `ClientCredentialsFlow` type for the endpoints instead of adding a new builder/constructor to the existing one. It seems pretty easy for the end user to dispatch to the right one at the time of building their application, as would already be the case when putting the refresh flow under the same `/token` endpoint, keyed off of `grant_type`.

For testing purposes, I added support for the client credentials grant to the actix version following the approach above, switching off `grant_type`. I tested it out manually in the console, it appears to be working. Once all looks good on the general approach, I will continue to add support for all the other frameworks and the async versions as well, and could add some buttons to the UI to make it easier to show in the example in future as well.

In particular I would like to confirm with you on these details as well:
- [x] For the client credentials flow, RFC-6749 section 4.4.3 states it `A refresh token SHOULD NOT be included` in the response. For now, this is just up to the issuer to determine that this is a client credentials grant, and not to issue a refresh token. Is this a good enough approach or should the flow indicate in some way that it is not recommended to include a refresh token, and require the end user to opt in (as they have to opt in to allow credentials in body)?
- [x] For the `owner_id` I have copied the `client_id` field as you mentioned, but is that necessarily the best option here? If the application is using ambiguous IDs, they could end up in a scary situation where a token appears to give access to a user's data when it was meant only for the client's data... but maybe that is just their problem?

Let me know if you have any other suggestions for this code. If this approach looks good I can finish up the rest of the framework integrations and find a way to improve the examples to show off this flow.

 - [x] Corresponds to issue #125


HeroicKatora (commented):
> For the client credentials flow, RFC-6749 section 4.4.3 states it A refresh token SHOULD NOT be included in the response.

I *believe* this recommendation to be based purely on usability: A refresh token is a long-lasting credential bound to the client. Using that refresh token will also require authenticating the client to the server. However, after such authentication of the client itself, there is nothing that the extra credentials of the refresh token would verify. It can just as well try to issue itself a new token with the same flow.

Nevertheless, the binding of the refresh token _could_ serve some business logic. For example, refreshing must only narrow the issued scope while a new flow could request any, potentially larger, scope. Also, use of the refresh token may be subject to different rate limits and access logging / resource links. So, it may be useful to just have it as an opt-in option. The RFC states `SHOULD NOT` which recognizes that it may be useful in some circumstances; but one should weight the full implications and use case against the recommendation. Opt-In seems appropriate, with the method to enable documenting the recommendation.

> If the application is using ambiguous IDs, they could end up in a scary situation where a token appears to give access to a user's data when it was meant only for the client's data... but maybe that is just their problem?

Is the concern that owners and client could be distinct entities, and this being a reinterpretation? Hm, that's not a bad argument. In the case of access token the respective string is produced by the `Solicitor` which is firmly under the implementors responsibilities. Maybe a similar approach should be taken here as well, as in: have a callback/customization point which assumes the role of translating client to owner—and a simple implementation that must only be used when they are indeed part of the same ID space.

If so, this should happen in the `Negotiate` phase of the flow. We could even reuse the same trait `Solicitor`, just with an internal failure if user-interaction is being required? That mechanism would be responsible for the implementor defined checks mentioned here:

> client is requesting access to the protected resources under its
   control, or those of another resource owner that have been previously
   arranged with the authorization server (the method of which is beyond
   the scope of this specification).
> -- RFC 6749 Section 4.4

foxfriends:
Thanks for your input, I have made those changes. There is now a flag `allow_refresh_token` that must be set, otherwise the refresh token will be explicitly excluded from responses. 

For the other issue, asking the solicitor is a good solution, so now the soliciting and then issuing steps occur after the state machine is done, as they do in `authorization` flow.

HeroicKatora (commented):
That seems to work rather nicely.

I've made one tiny high-level observation, mostly to have it written down somewhere: `Pending` is its own state machine and there's no big reason for this state machine to be separate from the main `code_grant` state machine. The `endpoint` is doing it separately mostly out of historical reasons as this was implemented before everything had been turned into Input/Output state machines. But please don't change anything based off this observation, it works properly, and has just slight code structure effects on `-async`.

I'm okay with accepting this as soon as it has a decently effective test suite. You can look at the [authorization test suite](https://github.com/HeroicKatora/oxide-auth/blob/master/oxide-auth/src/endpoint/tests/authorization.rs) for style and extent. If you identify any unique failure modes then please test them as well, signals from such regression testing have positively affected almost every single release.

foxfriends:
Sorry for the delay, busy weekend. I have added the tests, it seems to be ok.

The existing tests have some testing for "deny silently" but I realized that nothing I tried was causing that to happen. Is there some reasoning behind why certain requests should be denied silently, and others should be a 400/401 error? Seems to me every request I sent was failing 400...

HeroicKatora (approved):
'Silent' is in reference to a flow that uses redirects, and does not strongly authenticate the relying party before. As this can be triggered from *any* third-party by a mere link, certain errors shouldn't leak internal state: such as the existence of a particular client, or its allowed scopes for that particular user, etc. It's by no means a sufficient leak-freedom test suite but good to catch the basic errors.

None of this seems particularly relevant here. If anything it would be any form of information leak _before_ authenticating the client. I'll approve as is but since it's a large change I won't rush merging and publishing just yet. A few days of a fresh mind :)

foxfriends:
Ok sounds good to me, the async and other frameworks likely still need to be built out before publishing too right? I am using the async version personally, so I would like that before released anyway.

Let me know if you need any hand on those as well, I should have some time.

HeroicKatora:
> Sorry for the delay, busy weekend.

No worries, certainly. If anything, this project has clearly shown that slow-but-steady is more relevant than quick-and-dirty. We've sat at version `0.5.0` for the better of two years with a trickle of users and frameworks, and one large addition (async), and then one bug fix release. That's quite decent compatibility and reliability, if I do say so myself. It also left enough breathing room to write a migration document for 0.4–0.5 without burning out on the work. Just need to remember to come back every now and then.

> The async and other frameworks likely still need to be built out before publishing too right? I am using the async version personally, so I would like that before released anyway.

Not necessarily. It has an example with the core crate, that is convincing enough to warrant publishing `oxide-auth` itself. Incremental changes are fine here, as long as each step is sufficiently well motivated and designed. I'd say it's even a benefit, less risk to affect the whole. We can certainly do `oxide-auth-async in a separate PR.

oxide-auth/src/endpoint/tests/client_credentials.rs
```diff
@@ -0,0 +1,250 @@
...
+                ("client_id", "SomeOtherClient"),
+            ]
+            .iter()
+            .to_single_value_query(),
+        ),
+        auth: None,
+    };
+
+    ClientCredentialsSetup::new().test_bad_request(unknown_client, Allow("SomeOtherClient".to_owned()));
+}
```
HeroicKatora, line 196: It relates to the indistinguishabiltiy shortly touched upon before. During an authentication attempt it shouldn't be possible to distinguish between an unknown user and a bad password. Otherwise, an attacker might perform user enumeration by trying usernames.

This request, if done properly, should fail the same way as a bad password does (with `unauthenticated`). It currently doesn't because it's not actually sending a password—which is an invalid request. That property can be computed by the attacker so it's not a leak to respond this way. There should be a second test with a valid but unauthenticated other client id.

HeroicKatora (commented):
Letting it rest a bit was good. I found one more concern. I believe it is solved correctly but this test is misleadingly named. I can pick this up over the weekend myself.

PR #149, opened 2022-08-29
Modify async Issuer trait signature to mirror main crate

danielalvsaaker:
This changes the signature of `oxide_auth_async::primitives::Issuer` to mirror the signature of the related trait in the parent crate.

-----------------------------------------------------------


danielalvsaaker:
This actually doesn't seem doable without having to box/arc the issuer, since it's being passed around as `&dyn`.

PR #159, opened 2023-02-21
Implement an async version of client credentials flow

mtelahun:
This is an attempt to port client credentials flow to the async version. I noticed that it seemed to be the only implemented flow that was missing an async version. I didn't know if this was a deliberate omission or not, but after some investigation I realized that the async work was done much earlier. So, I decided to try my hand at porting it over. The work itself doesn't contain anything new. It's a straight up port of the non-synchronous code to an async version. Please let me know if I missed anything. Also, I would appreciate any comments on the minor changes in oxide-auth that are needed to get the async version to compile (like changing the visibility of a couple of functions and ::new() constructor in a couple of places).

 - [*] I have read the [contribution guidelines][Contributing]
 - [*] This change has tests (remove for doc only)
 - [*] This change has documentation

### DIFF
--- a/oxide-auth/src/code_grant/accesstoken.rs
+++ b/oxide-auth/src/code_grant/accesstoken.rs
@@ -617,7 +617,8 @@ impl Error {
 }
 
 impl PrimitiveError {
-    pub(crate) fn empty() -> Self {
+    /// Reset the results cache.
+    pub fn empty() -> Self {
         PrimitiveError {
             grant: None,
             extensions: None,
@@ -644,6 +645,11 @@ impl ErrorDescription {
 }
 
 impl BearerToken {
+    /// Given token parameters and a scope(s), create a new BearerToken.
+    pub fn new(token: IssuedToken, scope: String) -> BearerToken {
+        Self(token, scope)
+    }
+
     /// Convert the token into a json string, viable for being sent over a network with
     /// `application/json` encoding.
     pub fn to_json(&self) -> String {
--- a/oxide-auth/src/code_grant/refresh.rs
+++ b/oxide-auth/src/code_grant/refresh.rs
@@ -532,6 +532,11 @@ impl Error {
 }
 
 impl ErrorDescription {
+    /// Create a new description from an access token error
+    pub fn new(error: AccessTokenError) -> Self {
+        Self { error }
+    }
+
     /// Get a handle to the description the client will receive.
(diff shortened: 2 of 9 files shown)


oxide-auth/src/endpoint/mod.rs
```diff
@@ -121,7 +121,7 @@ pub enum ResponseStatus {
 /// not derive this until this has shown unlikely but strongly requested. Please open an issue if you
 /// think the pros or cons should be evaluated differently.
 #[derive(Debug)]
-enum InnerTemplate<'a> {
+pub enum InnerTemplate<'a> {
```
HeroicKatora, line 124: Can you add and use constructors to `Template` instead.

oxide-auth/src/code_grant/refresh.rs
```diff
@@ -532,6 +532,11 @@ impl Error {
 }
 
 impl ErrorDescription {
+    /// Create a new description from an access token error
+    pub fn new(error: AccessTokenError) -> Self {
+        Self { error }
+    }
```
HeroicKatora, line 538: Being critical towards making too many internals public applies here, too, the constructor should only be needed by the response template internally. There's already constructors applicable to `new_ok` and `new_unauthorized` for instance.

oxide-auth/src/code_grant/accesstoken.rs
```diff
@@ -644,6 +645,11 @@ impl ErrorDescription {
 }
 
 impl BearerToken {
+    /// Given token parameters and a scope(s), create a new BearerToken.
+    pub fn new(token: IssuedToken, scope: String) -> BearerToken {
+        Self(token, scope)
+    }
```
HeroicKatora, line 651: That should definitely not be the public signature, consider making the second parameter `&Scope`. But it's slightly controversial either way. This type consciously does not have a public constructor to nudge you towards properly using the flow and not issuing a badly protected token (e.g. to unauthenticated parties). I'll need to consider this, maybe you have arguments pro/con?

How complicated is it to duplicate? Seriously consider this alternative as well.

HeroicKatora (commented):
Reviewing the logic of the async state machine will take a short time and a breather, it being security critical. I like the suggestion and generic implementation though, just review the amount of items made public. Good encapsulation should not be lost.

mtelahun:
@HeroicKatora  Thanks for the feedback. I will update the PR as requested.

PR #161, opened 2023-04-11
improved scope struct access

Wicpar:
This improves the scope struct acces, by adding an `into_iter` and a safe `TryFrom<HashMap>`;

### DIFF
--- a/oxide-auth/src/primitives/scope.rs
+++ b/oxide-auth/src/primitives/scope.rs
@@ -2,6 +2,7 @@
 use std::{cmp, fmt, str};
 
 use std::collections::HashSet;
+use std::convert::TryFrom;
 use serde::{Deserialize, Serialize};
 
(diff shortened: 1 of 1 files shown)


HeroicKatora (commented):
`into_iter(self)` looks quite good, sure.

That said, I'm not sure if `TryFrom` is the right trait here, necessarily. In that the parameter `HashSet<String>` is quite specific and that specificity is motivated by implementation details primarily.

On the other hand, why not special case the zero-allocation conversion.

I'd be eager to explore alternatives and hear more arguments either.

Wicpar:
`TryFrom` can be implemented for any number of types on any given struct, there will never be any conflict unlike with `impl x for T`. Even if you wish to change the undelying structure you can always replace the implementation with `.into_iter().collect()` and implement the tryFrom for the new underlying structure.

PR #162, opened 2023-04-11
added serde to grant types

Wicpar:
This adds serde to the grant types

### DIFF
--- a/oxide-auth/src/primitives/grant.rs
+++ b/oxide-auth/src/primitives/grant.rs
@@ -7,6 +7,7 @@ use std::collections::HashMap;
 use std::collections::hash_map::Iter;
 use std::rc::Rc;
 use std::sync::Arc;
+use serde_derive::{Deserialize, Serialize};
 
 /// Provides a name registry for extensions.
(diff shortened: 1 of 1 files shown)


oxide-auth/src/primitives/grant.rs
```diff
@@ -24,7 +25,8 @@ pub trait GrantExtension {
 ///
 /// Some extensions have semantics where the presence alone is the stored data, so storing data
 /// is optional and storing no data is distinct from not attaching any extension instance at all.
-#[derive(Clone, Debug, PartialEq, Eq)]
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(tag = "type")]
```
HeroicKatora, line 29: For this type, there's a security concern when serializing `Private` since this may leak information rather unwittingly. I suppose it's okay overall but may be a paper cut when using the library. Just as a smoke-test, are there any alternatives that mitigate or avoid the risk?

HeroicKatora (commented):
Looks good, just a question in review.

PR #164, opened 2023-05-03
[FEATURE] Add client credential flow for oxide_auth_async

FrancisMurillo:
This adds the missing `client_credentials` flow for `oxide_auth_async`. 

For both the code and test, I just pretty much copied the sync `client_credential` and tweaked it according

Some possible issues with my code:
- I could not find a standard for `use` or imports
- `cargo clippy --fix` has a lot of fixes outside `oxide-auth-async`. I did not commit those and only for `oxide-auth-async`
- I had to use an `rust-toolchain.toml` set for the nightly channel so I think we should include that.
- I had to expose some internal types (changed `pub (crate)` to `pub`) which might need some attention

### DIFF
--- a/oxide-auth/src/code_grant/accesstoken.rs
+++ b/oxide-auth/src/code_grant/accesstoken.rs
@@ -563,13 +563,13 @@ pub struct PrimitiveError {
 /// addition this enforces backend specific behaviour for obtaining or handling the access error.
 #[derive(Clone)]
 pub struct ErrorDescription {
-    pub(crate) error: AccessTokenError,
+    pub error: AccessTokenError,
 }
 
 type Result<T> = std::result::Result<T, Error>;
 
 /// Represents an access token, a refresh token and the associated scope for serialization.
-pub struct BearerToken(pub(crate) IssuedToken, pub(crate) String);
+pub struct BearerToken(pub IssuedToken, pub String);
 
 impl Error {
     /// Create invalid error type
@@ -617,7 +617,7 @@ impl Error {
 }
 
 impl PrimitiveError {
-    pub(crate) fn empty() -> Self {
+    pub fn empty() -> Self {
         PrimitiveError {
             grant: None,
             extensions: None,
--- a/oxide-auth/src/code_grant/accesstoken.rs
+++ b/oxide-auth/src/code_grant/accesstoken.rs
@@ -12,6 +12,7 @@ use crate::primitives::authorizer::Authorizer;
 use crate::primitives::issuer::{IssuedToken, Issuer};
 use crate::primitives::grant::{Extensions, Grant};
 use crate::primitives::registrar::{Registrar, RegistrarError};
+use crate::primitives::scope::Scope;
 
 /// Token Response
 #[derive(Deserialize, Serialize)]
@@ -406,7 +407,7 @@ impl AccessToken {
     }
 
     fn finish(grant: Box<Grant>, token: IssuedToken) -> BearerToken {
-        BearerToken(token, grant.scope.to_string())
+        BearerToken(token, grant.scope.clone())
     }
 }
 
@@ -563,13 +564,13 @@ pub struct PrimitiveError {
 /// addition this enforces backend specific behaviour for obtaining or handling the access error.
 #[derive(Clone)]
 pub struct ErrorDescription {
-    pub error: AccessTokenError,
+    pub(crate) error: AccessTokenError,
 }
 
 type Result<T> = std::result::Result<T, Error>;
 
 /// Represents an access token, a refresh token and the associated scope for serialization.
-pub struct BearerToken(pub IssuedToken, pub String);
+pub struct BearerToken(pub(crate) IssuedToken, pub(crate) Scope);
 
 impl Error {
     /// Create invalid error type
@@ -617,6 +618,7 @@ impl Error {
 }
 
 impl PrimitiveError {
+    /// Create a default empty error
     pub fn empty() -> Self {
         PrimitiveError {
             grant: None,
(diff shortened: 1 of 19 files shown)


oxide-auth/src/endpoint/mod.rs
```diff
@@ -178,8 +178,8 @@ enum InnerTemplate<'a> {
 ///
 /// [`OwnerSolicitor`]: trait.OwnerSolicitor.html
 pub struct Solicitation<'flow> {
-    pub(crate) grant: Cow<'flow, PreGrant>,
-    pub(crate) state: Option<Cow<'flow, str>>,
+    pub grant: Cow<'flow, PreGrant>,
+    pub state: Option<Cow<'flow, str>>,
```
HeroicKatora, line 182: Here you can use `new` and `with_state` instead of making attributes public.

oxide-auth/src/code_grant/accesstoken.rs
```diff
@@ -563,13 +563,13 @@ pub struct PrimitiveError {
...
 pub struct ErrorDescription {
-    pub(crate) error: AccessTokenError,
+    pub error: AccessTokenError,
 }
 
 type Result<T> = std::result::Result<T, Error>;
 
 /// Represents an access token, a refresh token and the associated scope for serialization.
-pub struct BearerToken(pub(crate) IssuedToken, pub(crate) String);
+pub struct BearerToken(pub IssuedToken, pub String);
```
HeroicKatora, line 572: Making these have all-public attributes fully fixes their representation, which is not desirable. In particular, `String` should not appear in the interface. There could be a constructor from `IssuedToken` and `Scopes` instead. With a warning to only use scopes that come from a grant?

oxide-auth/src/code_grant/accesstoken.rs
```diff
@@ -563,13 +563,13 @@ pub struct PrimitiveError {
 /// addition this enforces backend specific behaviour for obtaining or handling the access error.
 #[derive(Clone)]
 pub struct ErrorDescription {
-    pub(crate) error: AccessTokenError,
+    pub error: AccessTokenError,
 }
```
HeroicKatora, line 567: This will disallow adding further fields to the description in compatible releases. Intentional?

HeroicKatora (commented):
I've not yet had time for the nitty gritty implementation of the state machine, so started reviewing the interface changes. That could be straightforward but maybe there's a difference in async?

HeroicKatora:
> cargo clippy --fix has a lot of fixes outside oxide-auth-async. I did not commit those and only for oxide-auth-async

Sounds good. Some of these clippy fixes are outdated code style, and the fact that turning off SemVer-impacting recommendations is an option that was only added recently so it's not configured correctly. This won't be an issue with this PR.

oxide-auth/src/endpoint/mod.rs
FrancisMurillo: Was able to use `Solicitation::new` instead

oxide-auth/src/code_grant/accesstoken.rs
FrancisMurillo: Huh, unintentional since reverting the change seems to have no impact on the tests.

FrancisMurillo: Intentional, needed for `client_credentials`. I just added a `ErrorDescription::new` function instead.

oxide-auth/src/code_grant/accesstoken.rs
FrancisMurillo: So is adding `IssuedToken::convert_bearer_token(self, PreGrant)` okay rather than accepting any `Scope`?

FrancisMurillo:
> > cargo clippy --fix has a lot of fixes outside oxide-auth-async. I did not commit those and only for oxide-auth-async
> 
> Sounds good. Some of these clippy fixes are outdated code style, and the fact that turning off SemVer-impacting recommendations is an option that was only added recently so it's not configured correctly. This won't be an issue with this PR. Thanks for investigating and commenting on it.

Applied the `cargo clippy` fix for all the packages now and applied some of the suggestions. There are some other warnings but I did not do those since they requires some deeper refactoring. 

One possible issue is I added some `.unwrap` to some `Result` that weren't handled. Most of them are in the test but just a heads up.

HeroicKatora:
> Applied the cargo clippy fix for all the packages now and applied some of the suggestions. There are some other warnings but I did not do those since they requires some deeper refactoring.

Please do not do this in the same PR. As you note, there are tiny differences which, as a reviewer, I'll have to search. Some of that is Github's fault by making it hard to approve PRs commit-by-commit. I'll ack <f8f90a5bbc8ce27820ee826b67e91c7f3df923c6> once rebased to that.

oxide-auth/src/endpoint/mod.rs
```diff
@@ -121,7 +121,7 @@ pub enum ResponseStatus {
 /// not derive this until this has shown unlikely but strongly requested. Please open an issue if you
 /// think the pros or cons should be evaluated differently.
 #[derive(Debug)]
-enum InnerTemplate<'a> {
```
HeroicKatora, line 124: I didn't catch that change in the first review. In general, I agree `From<_>` is a better constructor than the public methods duplicating all the variants on the type. However, the comment above explains the reasoning why it was chosen when the code was written, so please add `#[non_exhaustive]` and choose a better public name for the struct. Maybe `TemplateKind`, if you agree it is sufficiently clear (inspired by the split of  std::io::Error/std::io::ErrorKind)

HeroicKatora:
> Sorry, I thought you meant to apply it as well. Reverting it back.

Ah, yes, I see how that was easy to misinterpret. Sorry. No worries over this confusion.

oxide-auth/src/endpoint/mod.rs
FrancisMurillo: Not sure if I understand this correctly but I was able to revert the change from using `InnerTemplate` in favor of `Template` if that is okay.

HeroicKatora: That's fine as well :+1:

HeroicKatora (approved):

PR #165, opened 2023-05-11
[DRAFT] Add async/await support for `oxide-auth-async`

FrancisMurillo:

 - [ ] I have read the [contribution guidelines][Contributing]

Still fixing the main example work and adding another example to show explicit async/await usage


HeroicKatora:
On a general note: I'd actually rather replace the actix bindings with fully asynchronous ones rather than have two separate packages for it, if that sounds appropriate.

FrancisMurillo:
> On a general note: I'd actually rather replace the actix bindings with fully asynchronous ones rather than have two separate packages for it, if that sounds appropriate.

I tried doing in a separate crate first to see if there might be any issues with types and compatibility. I did have a problem with the actix actor example since the current implementation of `Generic` kinda needs to be `Clone`-able to pass the compiler check.

FrancisMurillo:
@HeroicKatora So this is kinda my async use case. (Mine was `diesel` but wrapped them in async domain methods.) I think you can inspect the current example while I try to make the original example with `Generic` work if possible.

ConcurrentSquared:
I don't want to bother anyone here (this PR is obviously inactive), but I do have a updated [fork of this PR](https://github.com/ConcurrentSquared/oxide-auth) with some very small improvements (the largest being the addition of the async `ClientCredentials` OAuth operation).

However, note that my fork is only for a personal side-project; I will almost certainly will **not** submit my own PR (or try to help with this PR), as I do not feel like I have either the time or experience (in Rust or in OAuth) to help in any significant way.

Issue #166, opened 2023-06-08
Future generated by `ClientCredentialsFlow` is not `Send`


aumetra:
# Bug report

The currently Git-only support for the client credential flow in `oxide-auth-async` produces a future that doesn't implement the `Send` auto-trait. This is due to a missing `Send` bound on a trait object.  

The effects of this bug are that the client credentials flow cannot be used in most web frameworks since most executors require the `Send` bound for their work stealing.

## Reproduction

Relevant environment:
- Axum (any web framework running on a workstealing executor, really)
- Work-stealing async executor

Steps to reproduce the behaviour:
1. Create a basic Axum-based server
2. Use the `ClientCredentialsFlow::execute` method in a handler and await the returned future
3. The project will fail to compile

## Expected behaviour

All other flows produce futures that implement the `Send` auto-trait, so it only makes sense for the client credentials flow to do the same.

Tracking pull request
--------

 - #167

PR #167, opened 2023-06-08
Add `Send` bound to trait object

aumetra:
This fixes the issue reported in #166 

 - [x] Corresponds to issue #166 


aumetra:
Regarding the second and third items in the list:
1. I can totally add a test that does a `Send` assertion on the futures returned by the flows.  
2. I'm not entirely sure what's to document with this change, the public API hasn't changed. The endpoint already had the `Send` requirement

HeroicKatora (commented):
Ah yes, this was definitely missed. If you can provide a simple test for that property at compile time, sure, go for it :slightly_smiling_face:  Regarding doc that's just a checklist, if there's nothing to do then there's nothing to do.

HeroicKatora (approved):

Issue #170, opened 2023-06-28
Implement `oxide_auth_async::endpoint::Enpoint<_>` for `oxide_auth::frontends::simple::extensions::Extended`


Geobert:
# Feature

While trying to use PKCE, I hit the fact that `oxide_auth::frontends::simple::extensions::Extended` does not implement `oxide_auth_async::endpoint::Enpoint<_>` (note the `async`)

## Alternatives

No alternative that I know of

## Tracking pull request

I’m willing to do this on behalf of my company but I need to discuss which path we should take.

As far as I understand the code, I should be able to implement the needed trait in `oxide_auth_async`, shouldn’t I?

Geobert:
Ok, not that simple as `inner` is private…

Geobert:
I have an issue though: When in access token flow, `Pkce::verify` doesn’t receive `method` as parameter (it’s `None`)

Geobert:
I don’t understand how the Extensions is supposed to go from AuthorizationFlow to AccessTokenFlow

HeroicKatora:
Do you have in-progress code? Maybe I can have a look and debug where the information is lost, and how to rework that data flow.

Geobert:
> Do you have in-progress code? Maybe I can have a look and debug where the information is lost, and how to rework that data flow.

I think I got it: I need to keep them in the login form to it is passed along ^^'

Are you alright with putting some field `pub` so we can implement the needed traits from oxide-auth-async?

(If yes, will be done in 2 weeks as I’m going in holidays today)

HeroicKatora:
I seem to be unable to spot any new commits in the branch you've linked. Generally, having the pass them in the form is potentially the intended path? But then again other applications would require storing them into the _server side_ token. Is there a reference or specification relating to the use case you're looking to solve?

Geobert:
I forgot to commit x) It’s done now :)

PR #171, opened 2023-07-07
fix: async trait impl for Extensions related stuff

Geobert:
This is needed to use PKCE in an async env

fixes https://github.com/HeroicKatora/oxide-auth/issues/170

### DIFF
--- a/oxide-auth-async/src/frontends/mod.rs
+++ b/oxide-auth-async/src/frontends/mod.rs
@@ -1,0 +1,1 @@
+pub mod simple;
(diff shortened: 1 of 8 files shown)


Geobert:
The cargo.toml is messed up (needed to remove the example as one of them needs nightly toolchain and forgot to restore them) Will do when I’m back ^^'

Geobert:
Hi, I’m back! 

Did you manage to have a look at this?

Geobert:
Is there anything wrong with this PR or it’s just a matter of having time to review it?

HeroicKatora:
Just a matter of having time to review. It's good you ask since it's slightly slipped my mind :slightly_smiling_face: 
Which is to say, any external review would also be appreciated even if it's a colleague of otherwise.

HeroicKatora (approved):
Probably needs a new release, right?

Geobert:
Thank you for reviewing/merging this!

Yes a release would be needed :)

PR #176, opened 2023-11-26
Disable axum's default features

aumetra:
This PR disables pulling in the default features of axum and instead only pulls in the two features it really needs (`query` and `form`)


HeroicKatora (approved):

danielalvsaaker:
LGTM! If you could also increment the minor version of the crate, then I can publish this change to crates.io.

aumetra:
Sure, I can increment it. 

But I have another breaking change (this time for the core `oxide-auth` crate), also around dependency count optimization (and maybe some dependency version updates)

So publishing to crates.io immediately wouldn't be optimal, I think (unless you don't mind two minor version bumps)

aumetra:
Oh, they actually just released axum v0.7. I guess that makes a good version bump

aumetra:
Rebased the PR

PR #177, opened 2023-11-27
Gate dependencies behind feature flags

aumetra:
This PR gates the following dependencies behind feature flags:

- `rust-argon2`
- `rmp-serde`

These dependencies are less common dependencies for some web services, since:

a) not everyone uses `rust-argon2` (personally, I use the RustCrypto `argon2` crate for pretty much all my projects)
b) not everyone uses MessagePack in their projects

In total, this makes the dependency tree 16 dependencies lighter with all features disabled.  
Whether these features should be enabled by default or not can definitely be discussed.

All the features are documented on docs.rs using the unstable `doc_auto_cfg` feature 

---


aumetra:
screenshot_docs

oxide-auth/src/endpoint/tests/pkce.rs
```diff
@@ -1,3 +1,5 @@
+#![cfg(feature = "client-map")]
```
HeroicKatora, line 1: Disabling the test suite is not a good implementation. We don't want to loose assurance due to a smaller feature selection. Can the dev-dependencies be extended accordingly to support this?

oxide-auth/src/primitives/generator.rs
```diff
@@ -72,152 +66,287 @@ impl RandomGenerator {
...
-/// signing the same grant for different uses, i.e. separating authorization from bearer grants and
-/// refresh tokens.
-pub struct Assertion {
-    hasher: Hmac<sha2::Sha256>,
-}
+#[cfg(feature = "assertion-grant")]
+pub use self::assertion_grant::{Assertion, AssertionKind, TaggedAssertion};
+
+#[cfg(feature = "assertion-grant")]
+mod assertion_grant {
```
HeroicKatora, line 73: Please use a separate file instead.

HeroicKatora (commented):
I don't agree with the approach. The two dependencies exist because some of the in-memory primitives come with default policies. However, the only effect of missing the dependencies could be that one must be provided instead of having a default / fallback. Indeed take `ClientMap` then it could expose a constructor with an explicit policy argument. Similarly the policy of using `rmp-serde` for the token signing could be boxed away as a trait that is provided in the constructor.

This allows the types to exist in principle, even when no default policies are found in the dependencies.

aumetra:
Makes sense. Will try to refactor towards generalizing the implementations over different inner implementations.

oxide-auth/src/primitives/generator.rs
aumetra: Done

oxide-auth/src/endpoint/tests/pkce.rs
aumetra: Generalized the `ClientMap` impl and this test now uses a `NoopPasswordPolicy` which just concatenates the Client ID and passphrase and checks for equality.  

Good enough for tests.

HeroicKatora:
The dispatch need not be byte-compatible with previous version, imo, since in-memory data should be treated as ephemeral. Well, any new design could be validated by trying to swap the current encoding scheme for something like BER or even JWT but those can always live as separate implementations as well. This opens the possibility of having a public wrapper type (with private fields, implementing Serialize) which the policy is to turn into a byte vector. Or rather two methods since there are two different instances of encoding happening. The `AssertGrant` type codifies essentially already one of those two, the other is currently ad-hoc tuples. Both could be public types, without public constructors.

aumetra:
Something like that? @HeroicKatora  

That's a fully opaque representation that publicly implements `Deserialize` and `Serialize` and doesn't expose any other information to the outside.  
Explicitly adds a disclaimer to the documentation that the internal representation is *unstable* and might change at any time.

The internal representation at the moment is just an enum.

HeroicKatora (commented):
Pretty much except that I'd prefer dynamic dispatch over a generic. It's a rather large operation where not a lot can be gained from an inline optimization. The only problem here are auto-traits, i.e. `Send` and `Sync` should still work for the `Assertion`. The inner types are all good :+1:

oxide-auth/src/primitives/generator/assertion_grant.rs
```diff
@@ -25,8 +59,9 @@ use super::TagGrant;
 /// The actual generator is given by a `TaggedAssertion` from `Assertion::tag` which enables
 /// signing the same grant for different uses, i.e. separating authorization from bearer grants and
 /// refresh tokens.
-pub struct Assertion {
+pub struct Assertion<E> {
     hasher: Hmac<sha2::Sha256>,
+    encoder: E,
```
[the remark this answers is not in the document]
aumetra: Dynamically dispatched the internals using an `Arc`. Added the required bounds (`Send + Sync + 'static`) to the trait itself to simplify all the generic methods that have to take a raw encoder.

aumetra:
> The only problem here are auto-traits, i.e. Send and Sync should still work for the Assertion

Rewrote the internals to dynamically dispatch instead. Works fine, added the auto traits as a requirement to the trait itself, so that's still fine.

HeroicKatora:
Fyi, the bulk of this PR looks good. But, from the perspective of _secure defaults_, I don't yet feel comfortable with an interface doesnt provide *some* secure configuration of all interfaces for users *by default*. That means, there should be an easy choice and that one should be secure. In particular, `new` with a parameter seems off. It's congruent with that idiom if the `new` interface is not available in all configurations so long as it is available in the default one.

The serializer is not as relevant but still I'm not sold on the idea of making it more difficult to most people (I'm basing this on the fact that this is the first time that issue came up) and not having at least an optional dependency that fully implements the trait. Or, at the *very least*, there should be good guidance within the trait documentation on how to bind it to a generic serialization library.

Just some thoughts here, and why I'm not convinced **yet** that this PR is ready.

aumetra:
@HeroicKatora Well, do you think it would be sufficient if we provided a `Default` implementation (even if gated behind a `cfg` flag) that uses these secure defaults?

Such as the interfaces that use the `argon2` password policy, they get a default impl.

And, of course, the `Encoder` also gets a default impl gated behind something like `messagepack`

aumetra:
@HeroicKatora Hey, sorry for the delay, I wanted to ask if what I proposed in the previous comment would be fine?  
Meaning an API where the default features expose these secure defaults?

HeroicKatora:
The approach is decent with the trait factoring, but I can't quite get over the _modification_ to `new` and other constructors. That introduces an apparently necessity for (or maybe draw to) an __improper__ implementation of the password store in tests. Knowing users, this might well be the place used as a template for customized versions. And if this is motivated by being 'just simple' then I can definitely see real cases of it happening in Production where it is a *huge* security hole. The changes you've made since the initial version do contribute to security-by-default but I'm not fully convinced that the PR overall is positive. It's close though.

PR #180, opened 2023-11-28
Upgrade dependencies in core crates

aumetra:
This PR updates outdated dependencies of the core crates. In particular, it updates:

- `base64`
- `rust-argon2`

Updating these since these are the ones that released semver-incompatible versions.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,14 @@
 # Changelog
 
-## [Unreleased]
+## `oxide-auth` [UNRELEASED]
+
+### Changed
+
+- Updated `base64` to v0.21
+- Updated `rust-argon2` to v2.0.0
+- The `Argon2` hasher now uses the parameters recommended by RFC-9106 for memory constrained environments
+
+## `oxide-auth-axum` v0.3.0
 
 ### Breaking 
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,3 +1,6 @@
+[profile.test]
+opt-level = 2
+
 [workspace]
 members = [
 	"oxide-auth",
--- a/oxide-auth-actix/Cargo.toml
+++ b/oxide-auth-actix/Cargo.toml
@@ -20,7 +20,7 @@ serde_urlencoded = "0.7"
 url = "2"
 
 [dev-dependencies]
-base64 = "0.13"
+base64 = "0.21"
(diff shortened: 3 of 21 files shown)


aumetra:
CI failure seems like a thing on the side of Cirrus CI, not a code issue

aumetra:
Actually this seems like an issue with the upgraded argon2 dependency. They adapted higher security parameters, requiring a bunch more memory and computation time.

That makes the tests time out.

I mean, we could either:

- leave the version as-is with lower defaults
- adapt the defaults of the RustCrypto argon2 crate, which seem reasonable but are much lower 
- make the argon2 parameters configurable from outside and using lower parameters for testing

Note that the parameters are so high, my laptop kills the test because of OOM 

The first option is rather undesirable since v2 fixed some UB around their parallel lane processing (they did mutable aliasing)

The second option would require us to adjust the parameters if they ever have to be readjusted (i.e. for security reasons)

The third option seems nice since it would let users adjust it themselves if the parameters are too high for their use case, but it would tie the public API to the `rust-argon2` crate, not letting us swap out argon2 implementations in the future. Unless we write a small wrapper struct.

HeroicKatora:
Would it already fix the performance if we ran all tests with `--release`? It seems that the recipe specifies debug, which means the machine code generated is likely in complete disagreement with the default parameters assumptions of the costs of compute and memory access.

aumetra:
> Would it already fix the performance if we ran all tests with --release? 

Let's try! I set the test profile to opt 2 which has a good performance/build time ratio for tests.

aumetra:
The Argon2 invocation now uses the low memory configuration as recommended by RFC9106.  
This is because the default configuration of `rust-argon2` (which are the recommended values by RFC9106), use 2GB of RAM to compute a hash.  
The low memory configuration uses ~65MB instead.

This is a tradeoff, however, since it technically compromises on the memory hardness of the function.  
But the somewhat more widely used `argon2` implementation by RustCrypto defaults to 19MB, which follows the [OWASP recommendation for minimal configurations](https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html#introduction).

So this seems like an acceptable tradeoff?

HeroicKatora (commented):
Agreed, sounds like decent reasoning. Would you add a line in the changenotes regarding that change to argon2, then this is definitely ready for merging.

HeroicKatora (approved):

PR #186, opened 2024-02-22
adjust OAuthResource to support usage

### LINKED ISSUE #185: oxide-auth-axum `OauthResource` deadtype
There are no examples for `oxide-auth-axum` (:shrug:) but it would appear that [`OAuthResource`](https://github.com/HeroicKatora/oxide-auth/blob/master/oxide-auth-axum/src/request.rs#L25) is a dead type. 

It may be constructed, but not used. 

Suggested improvement:
I am submitting a follow-up PR implementing:
```rust
impl OAuthResource {
    /// Fetch the authorization header from the request
    pub fn authorization_header(&self) -> Option<&str> {
        self.auth.as_deref()
    }
}
```

which will allow `OAuthResource` to be used to guard resources.

thor314:
This fixes #185 .

There are no examples for `oxide-auth-axum` (:shrug:) but it would appear that [`OAuthResource`](https://github.com/HeroicKatora/oxide-auth/blob/master/oxide-auth-axum/src/request.rs#L25) is a dead type. 

### DIFF
--- a/oxide-auth-axum/src/request.rs
+++ b/oxide-auth-axum/src/request.rs
@@ -133,3 +133,10 @@ where
         Ok(Self { auth })
     }
 }
+
+impl OAuthResource {
+    /// Fetch the authorization header from the request
+    pub fn authorization_header(&self) -> Option<&str> {
+        self.auth.as_deref()
+    }
+}


HeroicKatora:
It's not dead, there is a `From<_>` conversion to `OAuthRequest`. `axum` uses types in its handler signatures to route and for extractors, encouraging the use of different types for differing behavior. Here, `OAuthRequest` consumes the body while the simpler `OAuthResource` preserves it to be extracted in a subsequent extractor.

HeroicKatora (approved):

HeroicKatora:
Still approving this since there's no reason for requiring the conversion before inspecting the extracted header.

PR #190, opened 2024-05-21
fix: report request redirect uri instead of configured

### LINKED ISSUE #175: Issues with `IgnoreLocalPortUrl`
# Bug report

We try to use oxide-auth to test some OAuth clients in tests. During that, we encountered some issues regarding the usefulness of the current implementation of `IgnoreLocalPortUrl`.

### localhost vs. 127.0.0.1
https://www.rfc-editor.org/rfc/rfc8252.html#section-7.3 and https://datatracker.ietf.org/doc/html/draft-ietf-oauth-security-topics-23#section-4.1.3 call out to also support 127.0.0.1 for the special case of ignoring the port when validating the client-passed redirect_uri. As 127.0.0.1 (and probably the IPv6 localhost address) is considered to be safer as the hostname, I would expect that `IgnoreLocalPortUrl` also takes the loopback addresses.

### Completing a flow
It does not seem to be possible to complete a authorization code flow.
During `/authorization`, the redirect is generated by:
1. Using the `redirect_uri` of the `PreGrant`.
2. `PreGrant` (for ClientMap) gets the `redirect_uri` from `BoundClient` (`primitives/registrar.rs#negotiate`)
3. `BoundClient` (for ClientMap) gets the `redirect_uri` from the registered client (`primitives/registrar.rs#bound_redirect`)
This results in a browser being redirected to the wrong location.

## Reproduction

Relevant environment (http frontend library, OS, network setup?)

1. Using the actix example, setup the generic Endpoint using
```rs
IgnoreLocalPortUrl::new("http://localhost/endpoint")
   .unwrap()
   .into(),
```
2. Open http://localhost:8020/authorize?response_type=code&client_id=LocalClient&redirect_uri=http://localhost:8021/endpoint

## Expected behaviour

According to [rfc8252](https://www.rfc-editor.org/rfc/rfc8252.html#section-7.3), the redirect should happen based on the requests' `redirect_uri`.

One simple fix would be to change `bound_redirect` in registrar.rs to assign the following to `registered_url`
```rs
match registered {
   RegisteredUrl::IgnorePortOnLocalhost(_) => {
      RegisteredUrl::Exact(url.clone().into_owned())
   }
   url @ _ => url.clone(),
}
```

While this fixes the underlying issue, I am not sure of the consequences of this change. Having `RegisteredUrl` in `PreGrant` conveys, semantically by the type, that the content is the URL that is registered with the `Client`, but the field name and usage of the field suggests otherwise.

Tracking pull request
--------

 - [ ] A pull request does not yet exist

ctron:
Closes #175

 - [x] Corresponds to issue #175 

### DIFF
--- a/oxide-auth/src/primitives/registrar.rs
+++ b/oxide-auth/src/primitives/registrar.rs
@@ -787,14 +787,14 @@ impl Registrar for ClientMap {
         // Perform exact matching as motivated in the rfc
         let registered_url = match bound.redirect_uri {
             None => client.redirect_uri.clone(),
-            Some(ref url) => {
+            Some(url) => {
                 let original = std::iter::once(&client.redirect_uri);
                 let alternatives = client.additional_redirect_uris.iter();
-                if let Some(registered) = original
+                if original
                     .chain(alternatives)
-                    .find(|&registered| *registered == *url.as_ref())
+                    .any(|registered| *registered == *url.as_ref())
                 {
-                    registered.clone()
+                    RegisteredUrl::Exact((*url).clone())
                 } else {
                     return Err(RegistrarError::Unspecified);
                 }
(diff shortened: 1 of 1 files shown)


HeroicKatora (approved):
LGTM. Thank you!

Issue #191, opened 2024-05-21
Unable to find client for empty password


ctron:
# Bug report

The Rust `openid` crate sends the client ID and client secret as `username:password` via basic auth in the authorization header. However, when the `client_secret` is `None`, Reqwest's `basic_auth` function encodes this as `username:`.

Which gets decoded as `Some(b[])` by oxide-auth.

The effects of this bug are &#8230;

## Reproduction

Relevant environment (http frontend library, OS, network setup?)

1. Use `openid` on combination with a public client

## Expected behaviour

According to [reference], it should &#8230;

* As reqwest encodes this that way, I would expect this to be a proper way to convey "no password": https://github.com/seanmonstar/reqwest/blob/1f951d01bf3cda7d014698fd34adeba75706ac8c/src/util.rs#L15

I would also say that `Some([])` has the meaning of `None` in this case.

Tracking pull request
--------

 - [x] A pull request (does not yet exist/exists at #192)

ctron:
To my understanding of RFC6749 ([3.2.1](https://datatracker.ietf.org/doc/html/rfc6749#section-3.2.1).  Client Authentication) this would also be the correct behavior:

> A client MAY use the "client_id" request parameter to identify itself when sending requests to the token endpoint.  In the "authorization_code" "grant_type" request to the token endpoint, an unauthenticated client MUST send its "client_id" to prevent itself from inadvertently accepting a code intended for a client with a different "client_id".

ctron:
In the PR I went with `Option<Vec<u8>>` anyway. It feels more explicit about that it actually is. Let me know if you prefer it differently,

ctron:
Thanks for merging this. Would it be possible to do an new release with those changes?

PR #192, opened 2024-05-21
fix: if the password is empty, treat it as unauthenticated

ctron:

 - [x] Corresponds to issue (#191)


HeroicKatora (commented):
Probably yes :slightly_smiling_face: (going to sleep on the security relevance, but yes, seems like the right place for all the logic). Since those are concerned more with correct behavior it would be nice to see the full request setups represented in the 'integration test' portion of the test suite. The two sets of headers each are perfectly fine, just ensuring that the higher levels interpret missing passphrases as a public client etc.

https://github.com/HeroicKatora/oxide-auth/tree/master/oxide-auth/src/endpoint/tests

ctron:
Thanks for the pointer, I added some higher level tests for this.

HeroicKatora:
> In the PR I went with Option<Vec<u8>> anyway. It feels more explicit about that it actually is. Let me know if you prefer it differently.

> fn authorization(&self) -> Option<(Cow<str>, Option<Cow<[u8]>>)>;

I hadn't fully actualized that it touches public API. It's a bit by feel instead of a hard request, but the type alone would throw up some question marks if I'd encounter in trying to implement the trait. Could you try giving `Request::authorization` a public/custom enum to represent them as three options instead, see if that feels more intuitive to your or a colleague? 

I'm also thinking slightly beyond this PR, such a type might be quite suitable for other forms of authorization / tokens / by upgrading to `#[non_exhaustive]` or additional variants. Let's not rush that point, this is just for context.

HeroicKatora:
Also I'm comitting to that being the last requestion on the PR. I'm convinced of all the other details including that using the password this way is what's preferrential on all sides, if the enum outcome is unclear we can go ahead in the current state too.

HeroicKatora (approved):
My impression is, the enum pattern clarifies the options significantly.

Resolver is for the project workspace only and has no effect downstream, so yes, that's perfectly fine :+1:

PR #194, opened 2024-05-27
feat: implement Extension also for &mut AddonList

ctron:
There's a pattern to take a mutable reference of an endpoint and turning this into a new, modified endpoint.

This however doesn't work as some implementation for &mut AddonList are missing. This change implements those as well.

 - [x] Corresponds to issue (#195 )


ctron:
I also implemented `Clone` for the `AddonList`, which might come in handy in some cases.

ctron:
I added the fix for #195 to this PR as well. It seems to fit and actually builds on changes of this PR.

HeroicKatora (approved):
The nesting by implementing all the traits for `AddonList` seems very neat. Also interesting to see that `Clone` seems like a very natural derive that was missing.

Issue #217, opened 2026-01-22
Use Cargo example targets


ekohandel:
# Project Improvement

Currently the `oxide-auth-actix` and `oxide-auth-db` crates use a separate crate for their examples. As far as I know there is no reason for this and they could just use the example targets supported by cargo.

# Tracking pull request

 - [x] oxide-auth-actix - #218
 - [ ] oxide-auth-db

PR #218, opened 2026-01-22
example(oxide-auth-actix): make actix example part of oxide-auth-actix

ekohandel:
Move the actix example to be part of the oxide-auth-actix crate as an example instead of a separate binary crate of its own.

 - [x] Corresponds to #217 


ekohandel:
> The `path`-based module import becomes non-sensical in the published package. Plus all the dev-dependencies are heavy and I'm not sure I want them all linked in the (static) dependency tree for the package when they are motivated by the _frontend_ part, particular `reqwest`.
> 
> So, no, this separate crate is by design. It could use a `publish = false` key in its `Cargo.toml` however.

I see, I didn't know the original motivation for this. I will close this pull request and the associated issue then.

